PR #4, opened 2020-02-10
Tk add series mutation

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/SeriesFields.scala
```diff
@@ -1,19 +1,34 @@
...
       resolve = ctx => ctx.ctx.seriesService.getSeries(ctx.arg(BodyArg)),
       tags=List(ValidateBody()))
   )
+
+  val mutationFields: List[Field[ConsignmentApiContext, Unit]] = fields[ConsignmentApiContext, Unit](
+    Field(
+      "addSeries",
+      SeriesType,
+      arguments = List(SeriesInputArg),
+      resolve = ctx => ctx.ctx.seriesService.addSeries(ctx.arg(SeriesInputArg)))
```
suzannehamilton: What's the plan for authenticating and authorizing requests to this endpoint? If we're doing this properly, we should set up the Lambda/Jenkins job that will call this endpoint, and authenticate it with Keycloak.

My suggestion for populating the dropdown was to just add a migration with some fake seed data, and add the real population endpoint later.

TomJKing: Yes. There does need to be some authentication for the end point. I've asked Sam about what we should do for the authentication, ie a new role. OK. I thought we needed to add the endpoint to add the data. At least it will be ready for later.

MancunianSam: The way I've been writing it, there are two roles, `tdr_user` and `tdr_admin` The former being a normal user. 
I've added a commit https://github.com/nationalarchives/tdr-consignment-api/pull/4/commits/de562462d116e67056e166b9b319b19955f817f8 which adds a validator to check if the user is an admin user and rejects the request if they're not. There's a route unit test in there to show it working. 
We can have as many roles as we want within keycloak but for now I think user and admin are enough. If not, we can change the validation case class easily enough.

suzannehamilton: We'll eventually need at least one machine client role that the backend tasks will use, but we might not need it yet.

So is the plan for now that one of us will get an admin token, and then call this mutation from Postman (or similar)? How do we get a token?

I've added a card to remind us to write a proper background task when we come to add real series: https://github.com/orgs/nationalarchives/projects/20#card-32933151

TomJKing: Locally I was just printing the access token to console and using that in postman. That's probably not going to work so well in the proper environments thinking about it. Will write a database migration to add some seed data to the series and body table instead.

TomJKing: Seed migrations added for series and body tables: https://github.com/nationalarchives/tdr-consignment-api-data/tree/master/lambda/src/main/resources/db/migration

suzannehamilton:
We chatted about this in person yesterday. I'll try to summarise - let me know if I've misunderstood! :slightly_smiling_face: 

I'd prefer not to merge this PR, because we won't need this endpoint for a while. I think that it's generally better not to have code in the codebase that isn't being used, because you still have to maintain it (which is quite likely at this point in the project, while we're still deciding how we do things like structure the classes and write integration tests). It's also hard to maintain code that doesn't have a client, because we don't yet know exactly what the series updater background job will need to do. And it's the first admin endpoint, so we have to do make sure we're not introducing any security holes.

@TomJKing and @MancunianSam would prefer to merge this now because we're pretty certain that we'll need it soon. Merging it now is easier than merging it later, because the code is fresh in our minds (or at least in Tom's mind). The rest of the codebase will have changed by the time we merge it, so there might be lots of merge conflicts. Is that a fair summary?

So I think we just have a different instinct about the trade-offs. I'd rather we didn't merge it, but if Sam reviews it and you'd both rather merge it then that's OK. :slightly_smiling_face:

TomJKing:
Closed PR as per @suzannehamilton comments above

PR #6, opened 2020-02-17
Generate graphql schema

MancunianSam:
This adds a plugin to generate the graphql schema. This schema is then used to build the generated classes project which will be used by the APIs clients. I've also updated it to use the auth-utils project.


Jenkinsfile-deploy
```diff
@@ -13,9 +13,14 @@ pipeline {
             }
         }
         steps {
-            sh 'sbt -no-colors test scalastyle assembly'
+            sh 'sbt -no-colors graphqlSchemaGen scalastyle assembly'
```
suzannehamilton: Do we still want to run the `test` step?

MancunianSam: The assembly step runs the tests so we don't need it in there twice.

PR #15, opened 2020-02-20
Create consignment

MancunianSam:


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+    val mutation: String = fromResource(prefix + mutationPath).mkString
+    val expectedResult: String = fromResource(prefix + expectedResultPath).mkString
+    Post("/graphql").withEntity(ContentTypes.`application/json`, mutation) ~> addCredentials(token) ~> route ~> check {
+      print(responseAs[String])
+      responseAs[String] shouldEqual expectedResult
+    }
+  }
+
+  "The api" should "create a consignment if the correct information is provided" in {
+    runTestMutation("mutation_alldata.json", "data_all.json")
```
suzannehamilton: Are there any tests that check that the data is actually persisted, as well as checking the response?

src/test/resources/json/addconsignment_data_seriesid_missing.json
```diff
@@ -0,0 +1 @@
+{"data":null,"errors":[{"message":"Field 'AddConsignmentInput.seriesid' of required type 'Long!' was not provided. (line 1, column 47):\nmutation {addConsignment(addConsignmentInput: {userid: 1}) {consignmentid, seriesid, userid}}\n                                              ^","locations":[{"column":47,"line":1}]}]}
```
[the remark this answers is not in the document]
MancunianSam: You're right, the string comparison isn't great. I've put some changes in to decode the expected result and the response to an object and then compare those in the tests.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  "The api" should "throw an error if the series id field isn't provided" in {
+    runTestMutation("mutation_missingseriesid.json", "data_seriesid_missing.json")
+  }
+
+  "The api" should "throw an error if the user id field isn't provided" in {
+    runTestMutation("mutation_missinguserid.json", "data_userid_missing.json")
+  }
+
+  "The api" should "allow an tdr_admin user to create a consignment" in {
```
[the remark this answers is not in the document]
MancunianSam: I've changed the typo. I think it's ok for now, it's possible that in the future, admin users might create consignments on behalf of another user. Even if not, the roles are pretty basic at the moment and as you say, we need to think about them properly. Once we've decided what we're doing we can do it properly.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
MancunianSam: There weren't, I've added some in.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -0,0 +1,49 @@
...
+import sangria.schema.ScalarType._
+import sangria.macros.derive._
+import sangria.schema.{Argument, Field, InputObjectType, ListType, ObjectType, ScalarType, fields}
+import sangria.validation.ValueCoercionViolation
+import uk.gov.nationalarchives.tdr.api.graphql.ConsignmentApiContext
+
+import scala.util.{Failure, Success, Try}
+
+object ConsignmentFields {
+  private case object UuidCoercionViolation extends ValueCoercionViolation("Valid UUID expected")
```
TomJKing: May be split this out, as it is likely UUID are going to be used elsewhere for other objects

PR #22, opened 2020-03-17
Get transfer agreemement

MancunianSam:


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
```diff
@@ -90,6 +97,57 @@ class TransfersAgreementRouteSpec extends AnyFlatSpec with Matchers with TestReq
...
+  }
+
+  "The api" should "return an error if the user doesn't own the consignment" in {
+    val otherUserId = UUID.randomUUID().toString
+    val consignentSql = s"insert into consignmentapi.Consignment (SeriesId, UserId) VALUES (1,'$otherUserId')"
+    val ps: PreparedStatement = DbConnection.db.source.createConnection().prepareStatement(consignentSql)
+    ps.executeUpdate()
+    ps.executeUpdate()
+
+    val expectedResponse: GraphqlQueryData = expectedQueryResponse("data_invalidconsignment")
```
suzannehamilton: Should this be some kind of authorization error rather than "invalidconsignment"?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/Tags.scala
```diff
@@ -49,12 +51,18 @@ object Tags {
...
-  case class ValidateUserOwnsConsignment() extends ValidateTags {
+  case class ValidateUserOwnsConsignment(argName: String) extends ValidateTags {
     override def validate(ctx: Context[ConsignmentApiContext, _]): BeforeFieldResult[ConsignmentApiContext, Unit] = {
       val token = ctx.ctx.accessToken
       val userId = token.userId.getOrElse("")
-      val input: AddTransferAgreementInput = ctx.arg[AddTransferAgreementInput]("addTransferAgreementInput")
-      val result = ctx.ctx.consignmentService.getConsignment(input.consignmentId)
+
+      // Consignment id is either inside a case class extending UserOwnsConsignment or is a Long argument
+      val consignmentId: Long = Try(ctx.arg[UserOwnsConsignment](argName)) match {
```
suzannehamilton: I suspect this is going to get gnarly when we add more queries or mutations that need to share this validation but have a different set of arguments. This code is coupled in a non-obvious way to the code in TransferAgreementFields, and any future queries.

What about having separate tags and just sharing the common code, or extracting the relevant consignment ID in the Fields class, and passing it to this tag directly?

MancunianSam: What arguments are they going to have apart from a case class or a single Long value for the consignment id? This covers both cases.

Also, which part of this is coupled to the TransferAgreementFields?

I think having multiple validation tags with different names to do the exact same thing will be more difficult to maintain. It will take one dodgy copy and paste to have the wrong tag on a query looking for the wrong arguments. I know the tests should catch it but still.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
MancunianSam: I've changed the name to data_notowner because this is what we've used before and it makes more sense.

suzannehamilton: I might revisit this with the authorisation work so we can consolidate the auth errors, if that's OK?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/Tags.scala
suzannehamilton: Yeah, I think it's just a shame that Sangria doesn't make this particularly type-safe.

The thing that's a warning flag for me is that we're having to essentially do exception handling in this line, despite calling this method from a place where we know exactly which case we're in. If we had two separate tags, we wouldn't need the Try.

MancunianSam: True, I couldn't find a way to define a type bound that included both `Long` and `UserOwnsConsignment`
I think it's ok for now, if it does turn out to be an almighty mess then you get to say I told you so :smile:

suzannehamilton: OK, let's keep an eye on it getting more complex. We can refactor it two separate tags if this trips people up. 🙂

PR #24, opened 2020-03-19
Client file metadata mutation

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
```diff
@@ -0,0 +1,40 @@
...
+import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
+
+object ClientFileMetadataFields {
+  case class ClientFileMetadata(fileId: Long,
+                                originalPath: Option[String] = None,
+                                checksum: Option[String] = None,
+                                checksumType: Option[String] = None,
+                                lastModified: Long,
+                                createdDate: Long,
+                                fileSize: Option[BigDecimal] = None,
```
suzannehamilton: Are some of these fields optional because they're nullable in the DB? Do you think they should be made non-nullable? I can't think of a scenario where we'd have some of these fields but not all of them.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
```diff
@@ -0,0 +1,40 @@
...
+import sangria.marshalling.circe._
+import sangria.schema.{Argument, Field, InputObjectType, ObjectType, fields}
+import uk.gov.nationalarchives.tdr.api.graphql.ConsignmentApiContext
+import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
+
+object ClientFileMetadataFields {
+  case class ClientFileMetadata(fileId: Long,
+                                originalPath: Option[String] = None,
+                                checksum: Option[String] = None,
+                                checksumType: Option[String] = None,
```
suzannehamilton: This reminds me: do you know what checksum algorithms we're going to use in the frontend? Are we planning to just use one algorithm? Just wondering if this DB column still makes sense.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ClientFileMetadataRouteSpec.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  private def checkClientFileMetadataExists(clientFileMetadataId: Long): Unit = {
+    val sql = s"select * from consignmentapi.ClientFileMetadata where ClientFileMetadataId = $clientFileMetadataId;"
+    val ps: PreparedStatement = DbConnection.db.source.createConnection().prepareStatement(sql)
+    val rs: ResultSet = ps.executeQuery()
+    rs.next()
+    rs.getString("ClientFileMetadataId") should equal(clientFileMetadataId.toString)
+  }
+
+  private def resetDatabase(): Unit = {
```
suzannehamilton: Are we doing DB resets in a lot of places now, and would it be easier to share the teardown by wiping the DB after every test?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
TomJKing: I was just following what is in the current DB schema. Is it not a possibility that something might fail for a file, for example filesize but other data doesn't, having the fields nullable would still let data to be stored in the DB, if that is desirable

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
TomJKing: No idea. Is it likely we would change checksum algorithm at some point? If so this column would still make sense

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ClientFileMetadataRouteSpec.scala
TomJKing: There are a lot of resets, but I'm not sure having a sharedown is a good idea. From a personal point of view with unit tests I prefer everything in one place. I've seen having a shared functionality between tests causing unintended consequences across unit tests

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
suzannehamilton: Hmm, I suspect it's not desirable, and it's not error handling we were planning to write any time soon, but we probably have to think about the options.

I might run this one past David and Kirsten to see what data is essential. My guess would be that it's better to fail the upload than save a file without a client-side checksum, but I'll check with them and create a new card if necessary.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ClientFileMetadataFields.scala
suzannehamilton: Sorry, I meant that if we were using more than one algorithm then we'd want to structure the data differently.

I'll check with David to make sure he's still happy with the algorithm we used in the prototype.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ClientFileMetadataRouteSpec.scala
suzannehamilton: To be honest, it would be nice if we could create and tear down the entire DB for each test (and not have many integration tests, because that's slow). I'm definitely not suggesting that we have tests that depend on setup from previous tests! 🙂 

I'm actually doing some API dev at the moment, so it's nice to get a feel for how easy the tests are to work with. Let's leave it like this for now. I might try refactoring things a bit if anything seem fiddly to work with, and then see what you all think.

PR #27, opened 2020-03-19
Add line break to end of file

TomJKing:
Missing line break causing build to fail


suzannehamilton:
Merging this to fix the build.

PR #28, opened 2020-03-19
Get consignment creator from token

suzannehamilton:
The user's ID is used for auditing purposes, so we should read it from the user's authentication token rather than trusting the UUID sent by the client.

For now, make the user ID parameter in the GraphQL request optional, so that this change is backwards-compatible with the frontend client. The frontend will be updated to remove the parameter, and then the parameter can be removed from the API.

### DIFF
--- a/src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
+++ b/src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
@@ -51,12 +51,6 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
     response.errors.head.message should equal (expectedResponse.errors.head.message)
   }
 
-  "The api" should "throw an error if the user id field isn't provided" in {
-    val expectedResponse: GraphqlMutationData = expectedMutationResponse("data_userid_missing")
-    val response: GraphqlMutationData = runTestMutation("mutation_missinguserid", validUserToken())
-    response.errors.head.message should equal (expectedResponse.errors.head.message)
-  }
-
   "The api" should "allow a tdr_user user to create a consignment" in {
     val expectedResponse: GraphqlMutationData = expectedMutationResponse("data_all")
     val response: GraphqlMutationData = runTestMutation("mutation_alldata", validUserToken())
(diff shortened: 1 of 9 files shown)


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -39,38 +39,32 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
...
-    val response: GraphqlMutationData = runTestMutation("mutation_missingseriesid", validAdminToken)
-    response.errors.head.message should equal (expectedResponse.errors.head.message)
-  }
-
-  "The api" should "throw an error if the user id field isn't provided" in {
-    val expectedResponse: GraphqlMutationData = expectedMutationResponse("data_userid_missing")
-    val response: GraphqlMutationData = runTestMutation("mutation_missinguserid", validAdminToken)
+    val response: GraphqlMutationData = runTestMutation("mutation_missingseriesid", validUserToken())
     response.errors.head.message should equal (expectedResponse.errors.head.message)
   }
```
MancunianSam: I think we still need the test to make sure that an admin user can create a consignment as well.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/FixedTimeSource.scala
```diff
@@ -0,0 +1,9 @@
+package uk.gov.nationalarchives.tdr.api.utils
+
+import java.time.{Instant, LocalDateTime, ZoneOffset}
+
```
MancunianSam: Couple of unused imports from java.time here.

suzannehamilton: Fixed, thanks.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
MancunianSam: Sorry, I've just read the commit comment, ignore me.

suzannehamilton: I considered doing that, but skipped it in the end because I think we're going to have to think about what types of users we need in TDR. I don't think there's any admin user functionality in the MVP plan. There will be backend clients for things like the antivirus check or the series import, but they're not really admin users either - they'll need their own role and restricted permissions.

suzannehamilton: Sorry, I posted before I saw your comment! 🙂

PR #29, opened 2020-03-20
Remove user from consignment creation query

suzannehamilton:
In a previous commit, we switched to getting the user ID from the auth token, so it is no longer needed in the GraphQL query.

PR marked as draft until https://github.com/nationalarchives/tdr-transfer-frontend/pull/58 is merged.

Edit: I found an issue in the graphql client that needs to be fixed first too: https://github.com/nationalarchives/tdr-graphql-client/pull/2


suzannehamilton:
All the dependencies are merged, so this one is ready to be reviewed and merged.

PR #30, opened 2020-03-23
Add file

MancunianSam:
This was originally going to be an endpoint for adding a single file but the fields on the File table are:
FileId,
ConsignmentId,
UserId,
DateTime

Only the consignment id isn't either auto-generated or got from the token, there's no information about the actual file here, that's all stored in the ClientFileMetadata table.
So it makes very little sense to have an endpoint that only creates one file and just calling it with the same arguments over and over for each file. So I've changed it to work like this.


suzannehamilton:
Thought which should have occurred to me in the planning meeting: does it even make sense to have separate endpoints for creating a file and adding client-side metadata? If the client-side metadata is the first information we have about a file, should the file and the metadata be created at the same time?

We shouldn't have to structure our API endpoints in exactly the same way as the DB tables, because that implies we can't change the DB structure later without changing the API.

MancunianSam:
I think it is worth it. The flow in the front end would be like this:

* User clicks upload
* We call this mutation which populates the database with all of the files we're expecting to receive and get a list of ids back.
* We get the client metadata from the library
* We can then call the client metadata mutation using the ids returned from the first step. There's no need to match the returned file ids to the individual files because the individual files contain no information about the file itself. 

If the upload fails for whatever reason or the user closes their browser, we have a records of how many files we're expecting to get which will help with error handling later. 

If we have a joint file and metadata endpoint and only call the mutation when we have the metadata back from the library, we'll have no idea how many files we were expecting to get.

MancunianSam:
The main reason we need to know how many files we have as soon as possible is for the backend checks. If we only get half the files in S3 because they've closed their browser and we don't know how many files we were expecting, the backend checks will show 100% complete when in fact we're missing a load of files.

suzannehamilton:
Won't we get that from the endpoint we're planning to add later, which just posts the total number of files in the consignment?

suzannehamilton:
Sorry, have I misunderstood, and this _is_ the endpoint which takes the number of files?

MancunianSam:
Yeah, it also gives us the advantage of having the file ids in the front end which we can then use to post the client metadata.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
```diff
@@ -0,0 +1,30 @@
...
+import io.circe.generic.auto._
+import sangria.macros.derive._
+import sangria.marshalling.circe._
+import sangria.schema.{Argument, Field, InputObjectType, ObjectType, fields}
+import uk.gov.nationalarchives.tdr.api.graphql.ConsignmentApiContext
+import uk.gov.nationalarchives.tdr.api.graphql.Tags.ValidateUserOwnsConsignment
+import uk.gov.nationalarchives.tdr.api.graphql.validation.UserOwnsConsignment
+
+object FileFields {
+  case class File(fileIds: Seq[Long])
```
suzannehamilton: Are file IDs going to be UUIDs rather than longs, like in the prototype? The biggest advantage is that we'll have guaranteed unique IDs, even if we delete data from the DB (e.g. in integration/staging).

suzannehamilton:
It would be good to have a chat about this - I've sent you a message on Slack. 🙂

suzannehamilton:
As discussed:

- I'll create a card for switching to Postgres, so we can use UUID IDs
- We'll stick with this file creation flow for now, but we need to see how it performs in stress-testing, because the API response will include every file ID even if there are thousands of files. Will the API/DB/browser cope?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
suzannehamilton: As discussed, we're going to switch to Postgres (which was the original choice of DB) and then use UUID IDs.

Just to clarify something @MancunianSam: did you say MySQL doesn't support UUID primary keys, or is it more complicated than that? Because I've seen some docs saying that it is possible ([example](https://www.mysqltutorial.org/mysql-uuid/)).

src/test/resources/json/addfile_mutation_alldata.json
```diff
@@ -0,0 +1 @@
+{"query":"mutation {addFiles(addFilesInput: {consignmentId: 1, numberOfFiles: 3}) {fileIds}}"}
```
suzannehamilton: Something for later, but I was wondering whether Jenkins should validate all these test queries against the schema? So we always know we're running real, valid queries in the tests?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+  "createFile" should "create three files given correct arguments" in {
+    val uuid = UUID.randomUUID()
+    val consignmentId = 1
+    val fileRepositoryMock = mock[FileRepository]
+    val fileRowOne = FileRow(consignmentId, uuid.toString, Timestamp.from(Instant.now), Some(1))
+    val fileRowTwo = FileRow(consignmentId, uuid.toString, Timestamp.from(Instant.now), Some(2))
+    val fileRowThree = FileRow(consignmentId, uuid.toString, Timestamp.from(Instant.now), Some(3))
+    val mockResponse = Future.successful(List(fileRowOne, fileRowTwo, fileRowThree))
+
+    when(fileRepositoryMock.addFiles(any[List[FileRow]])).thenReturn(mockResponse)
```
suzannehamilton: In this case, where the service actually has some logic, I think we should be stricter about what we're testing. For example, we should make sure that it creates the number of rows being passed in, not just any list of FileRows.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+    val fileService = new FileService(fileRepositoryMock, FixedTimeSource)
+    val result: File = fileService.addFile(AddFilesInput(consignmentId, Some(3)),Some(uuid)).await()
+
+    result.fileIds shouldBe List(1,2,3)
+  }
+
+  "createConsignment" should "link a consignment to the user's ID" in {
```
suzannehamilton: Copy-pasted test name?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
MancunianSam: I didn't realise that. I thought that mysql didn't support them. Didn't you say that Postgres supports some other things which we need that mysql doesn't though?

suzannehamilton: Here's one example: https://github.com/nationalarchives/tdr-consignment-api-data/pull/13

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -0,0 +1,21 @@
...
+import uk.gov.nationalarchives.Tables.FileRow
+import uk.gov.nationalarchives.tdr.api.db.repository.FileRepository
+import uk.gov.nationalarchives.tdr.api.graphql.fields.FileFields.{AddFilesInput, File}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class FileService(fileRepository: FileRepository, timeSource: TimeSource)(implicit val executionContext: ExecutionContext) {
+
+  def addFile(addFilesInput: AddFilesInput, userId: Option[UUID]): Future[File] = {
+    val rows: Seq[nationalarchives.Tables.FileRow] = List.fill(addFilesInput.numberOfFiles.getOrElse(1))(1)
```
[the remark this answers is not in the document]
MancunianSam: I'm not sure what I was thinking there, I think you're right in the comment above and it just needs to be a mandatory parameter.

src/test/resources/json/addfile_mutation_alldata.json
MancunianSam: We're running these queries against the real endpoints though so if they're invalid, the tests will fail. We do have some tests that remove mandatory parameters and check for the schema errors.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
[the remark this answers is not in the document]
MancunianSam: Files makes more sense, I've change it.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  case class AddFilesInput(consignmentId: Long, numberOfFiles: Option[Int]) extends UserOwnsConsignment
```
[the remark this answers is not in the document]
MancunianSam: Yeah, you're right, I've made it mandatory and updated/deleted the tests as appropriate.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
MancunianSam: Ah, fair enough, that's one we can leave out when we move to postgres.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: I've added in some argument captors so we can check which arguments were passed to the repository mock.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: Yep, cheers

build.sbt
```diff
@@ -39,7 +39,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-optics" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
-  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.11",
+  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.12-SNAPSHOT",
```
TomJKing: Should this be the snapshot version?

src/test/resources/scripts/init.sql
```diff
@@ -50,5 +50,14 @@ CREATE TABLE IF NOT EXISTS consignmentapi.ClientFileMetadata (
    PRIMARY KEY (ClientFileMetadataId)
 );
 
+CREATE TABLE IF NOT EXISTS consignmentapi.File (
+   FileId bigint(20) NOT NULL AUTO_INCREMENT,
+   ConsignmentId bigint(20) NOT NULL,
+   UserId varchar(40) DEFAULT NULL,
+   Datetime timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+   PRIMARY KEY (FileId)
+
```
TomJKing: Unnecessary line break

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private def checkFileExists(fileId: Long) = {
+    val sql = s"select * from consignmentapi.File where FileId = $fileId;"
+    val ps: PreparedStatement = DbConnection.db.source.createConnection().prepareStatement(sql)
+    val rs: ResultSet = ps.executeQuery()
+    rs.next()
+    rs.getString("FileId") should equal(fileId.toString)
+  }
+
```
TomJKing: Unnecessary line breaks

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
suzannehamilton: The second bit still looks like it's from the consignment test.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
suzannehamilton: I see the captors in the test below. Maybe you could add them here too, to cover the case where there's more than one file?

build.sbt
MancunianSam: Good spot, cheers.

src/test/resources/scripts/init.sql
MancunianSam: I always like to leave you one :smile:

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
MancunianSam: Or at least one anyway.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: Sorry, got my line numbers confused. I've added it in here as well.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: I've updated it again and change the argument captor to check for the consignment id and user id.

PR #32, opened 2020-03-25
Make transferring body a compulsory argument when fetching series

suzannehamilton:
This fixes an authorisation issue, where it was possible to fetch series for all departments by passing a null body parameter.

The only reason that `body` was an optional parameter was to support future admin functionality, but that does not exist yet so it's more secure to make the parameter compulsory are revisit this once we have requirements for the admin functions.

Also remove the admin token handling the same endpoint for the same reason.


suzannehamilton:
Closing because I meant to mark this as not mergeable yet.

PR #33, opened 2020-03-25
Make transferring body a compulsory argument when fetching series

suzannehamilton:
This fixes an authorisation issue, where it was possible to fetch series for all departments by passing a null body parameter.

The only reason that body was an optional parameter was to support future admin functionality, but that does not exist yet so it's more secure to make the parameter compulsory are revisit this once we have requirements for the admin functions.

Also remove the admin token handling the same endpoint for the same reason.

Set as draft for now. The frontend needs to be updated to use https://github.com/nationalarchives/tdr-generated-graphql/pull/11 before this can be safely merged.


suzannehamilton:
Rebased to fix the merge conflicts.

This can now be merged because the frontend is on the latest version of the GraphQL library.

PR #36, opened 2020-03-25
Add error code to GraphQL errors

suzannehamilton:
GraphQL supports error codes to provide more information to the client applications. We can look for these code Strings in the GraphQL clients to provide better errors to the user.

For now, the only specialised error is NOT_AUTHORISED, which indicates that the request tried to do something which the user does not have permission for.


src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -25,7 +25,11 @@ import scala.util.{Failure, Success}
 object GraphQLServer {
 
   val exceptionHandler = ExceptionHandler {
```
suzannehamilton: Something else we need to do is to log the errors caught here so that we don't throw away the original exception information.

We haven't set up any logging yet, so I might add a card to the backlog rather than doing it as part of my current card.

PR #38, opened 2020-04-03
Refactor authorisation tags

suzannehamilton:
This is probably easier to review commit-by-commit, because I've moved the Tags file as well as refactoring.

All of the commits are refactoring to make the authorisation easier. There are no changes to the functionality. I just thought it would be easier to review it separately, rather than including it with the new auth checks.

I've refactored a few things

- Move the authorisation tags to the auth package, to distinguish them from any future non-auth tags
- Extracting the `Await.result` step to the middleware, so that I don't have to duplicate it when adding new authorisation tags. I've raised a card about fixing this properly (see [TDR-52](https://national-archives.atlassian.net/browse/TDR-52)), but that will be easier to do if it's isolated in one place.
- A bit of Scala tidy-up: making tag classes traits or objects where possible


suzannehamilton:
I've just rebased to fix the merge conflicts with the Long -> UUID conversion. I'll merge if the Jenkins build passes.

PR #41, opened 2020-04-06
Ensure all endpoints have authorisation

suzannehamilton:
- Add authorisation to all queries and mutations that didn't have it, to make sure clients have permission to call that endpoint
- Add a check to the ValidationAuthoriser to catch any top-level fields which don't have authorisation


src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -59,18 +59,24 @@ object GraphQLServer {
     val uuidSourceClass: Class[_] = Class.forName(ConfigFactory.load().getString("source.uuid"))
     val uuidSource: UUIDSource = uuidSourceClass.getDeclaredConstructor().newInstance().asInstanceOf[UUIDSource]
     val db = DbConnection.db
+
+    val consignmentRepository = new ConsignmentRepository(db)
```
TomJKing, line 63: Is there a reason why the consignment repo object has been assigned to a val, unlike the other repos which are defined in the service constructors?

suzannehamilton: It's because it's being used more than once - it's passed to the ConsignmentService and the FileService.

I couldn't find a neater way to do it, but I thought it was OK on balance - some of the services might need to fetch data from different repositories in order to keep some business logic in the service layer, so it's not cluttering the the GraphQL layer. What do you think?

suzannehamilton:
There are unit tests for the TransferringBodyService but no route spec because there aren't currently any queries or mutations for transferring bodies. The TransferringBodyService is used in the authorisation for one of the series fields.

Does that answer your question, or have I missed some tests that should be there?

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
TomJKing: OK. Missed it being used more than once. Seems like a reasonable approach.

TomJKing:
Yes. That answers my question. Looks OK

suzannehamilton:
@TomJKing There were quite a lot of merge conflicts to fix, so would you mind re-reviewing this one commit? https://github.com/nationalarchives/tdr-consignment-api/pull/41/commits/8599e6637d5fc4c9792460e855a96bd1f3617351

I've updated it to handle the changes to the addClientSideMetadata mutation. Now that it takes a list of files rather than a single file, I've updated the authorisation to check the ownership of all the files.

PR #43, opened 2020-04-22
Prevent multiple uploads to same consignment

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -119,3 +119,19 @@ object ValidateUserOwnsFiles extends AuthorisationTag {
...
+  override def validate(ctx: Context[ConsignmentApiContext, _])
+                       (implicit executionContext: ExecutionContext): Future[BeforeFieldResult[ConsignmentApiContext, Unit]] = {
+    val arg: T = ctx.arg[T](argument.name)
+    val consignmentId: UUID = arg match {
+      case uoc: UserOwnsConsignment => uoc.consignmentId
+      case id: UUID => id
+    }
+
+    ctx.ctx.consignmentService.previousUploadToConsignment(consignmentId).map {
+      case true => throw AuthorisationException("Upload already occurred for consignment: " + consignmentId)
```
suzannehamilton, line 133: This is a new type of error - it's more about the consignment not being in the right state rather than an authorisation problem. (i.e. the user is allowed to upload files to this consignment - it's just the wrong stage in the transfer process for it.)

Maybe we should make a new file for tags that check consignment state? And then add a different type of exception, and a different exception representing invalid state? You can give that its own error code in the `exceptionHandler` in GraphQLServer.scala.

If possible, we should get Sangria to check the authorisation tags before the state tags, because if a user isn't allowed to do a certain action then we should tell them that rather than give them information about the consignment that they're not supposed to see.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -86,6 +86,19 @@ class FileRouteSpec extends AnyFlatSpec with Matchers with TestRequest with Befo
...
+  "The api" should "throw an error if the consignment already has had files uploaded" in {
+    val sqlConsignment = s"insert into Consignment (SeriesId, UserId) VALUES (1,'$userId')"
+    val psConsignment: PreparedStatement = DbConnection.db.source.createConnection().prepareStatement(sqlConsignment)
+    psConsignment.executeUpdate()
+    //Seed DB with initial file for consignment
+    runTestMutation("mutation_one_file", validUserToken())
+
+    val expectedResponse: GraphqlMutationData = expectedMutationResponse("data_error_previous_upload")
+    val response: GraphqlMutationData = runTestMutation("mutation_one_file", validUserToken())
+    response.errors.head.message should equal (expectedResponse.errors.head.message)
```
suzannehamilton, line 98: Once you've added a new GraphQL error code, it would be good to check that here too.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -23,4 +23,10 @@ class ConsignmentService(consignmentRepository: ConsignmentRepository, timeSourc
 
     consignments.map(rows => rows.headOption.map(row => Consignment(Some(row.consignmentid), row.userid, row.seriesid)))
   }
+
+  def previousUploadToConsignment(consignmentId: UUID): Future[Boolean] = {
```
suzannehamilton, line 27: Minor thing, but I might be inclined to word this more generically, e.g. `consignmentHasFiles`, because the service doesn't need to know that you're checking a "previous" upload against a current one.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -21,6 +21,14 @@ class ConsignmentRepository(db: Database) {
     db.run(query.result)
   }
 
+  def getConsignmentFiles(consignmentId: UUID): Future[Seq[UUID]] = {
+    val query = for {
+      (file, _) <- File.join(Consignment).on(_.consignmentid === _.consignmentid).filter(_._1.consignmentid === consignmentId)
```
suzannehamilton, line 26: In answer to your question about whether you need the join - I suspect not, because you don't need anything from the consignment table.

I guess we really only need to know if there are any files for the consignment. I was going to suggest the Slick equivalent of a SQL `COUNT` query, but I've just found the `exists` function: https://stackoverflow.com/a/33628053/679924 Does that do what you need, since it avoids having a query that might return data for thousands of files?

TomJKing: Yes, that's a nice function. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
TomJKing: Thanks. I was wondering about this, as the validation checks for the consignment didn't really fit under authorisation checks.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
TomJKing: Fair point.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
TomJKing: Added test for the new error code

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationAuthoriser.scala
```diff
@@ -25,13 +25,13 @@ class ValidationAuthoriser(implicit executionContext: ExecutionContext)
...
     val isTopLevelField = ctx.path.path.length == 1
-    val fieldHasAuthTag = ctx.field.tags.exists(tag => tag.isInstanceOf[AuthorisationTag])
+    val fieldHasAuthTag = ctx.field.tags.exists(tag => tag.isInstanceOf[ValidationTag])
     if (isTopLevelField && !fieldHasAuthTag) {
       throw new AssertionError(s"Query '${ctx.field.name}' does not have any authorisation steps defined")
     }
 
     val validationList: Seq[BeforeFieldResult[ConsignmentApiContext, Unit]] = ctx.field.tags.map {
-      case v: AuthorisationTag => {
+      case v: ValidationTag => {
```
suzannehamilton, line 34: Is this check strict enough now? The goal of this check was to make sure every API endpoint had some authorisation defined. It's not enough for it to have state validation but not authorisation.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationTag.scala
```diff
@@ -0,0 +1,14 @@
+package uk.gov.nationalarchives.tdr.api.auth
```
suzannehamilton, line 1: It would be good to move this out of the auth package if it's being used for more than just auth.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/ErrorCodes.scala
```diff
@@ -8,4 +8,5 @@ package uk.gov.nationalarchives.tdr.api.graphql
   */
 object ErrorCodes {
   val notAuthorised = "NOT_AUTHORISED"
+  val invalidConsignmentState = "INVALID_CONSIGNMENT_STATE"
```
suzannehamilton, line 11: 👍

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -88,4 +88,34 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
...
+    val fixedUuidSource = new FixedUUIDSource()
+    val fileUuid1 = UUID.randomUUID()
+    val fileUuid2 = UUID.randomUUID()
+    val mockResponse: Future[Boolean] = Future.successful(true)
+    val consignmentRepoMock = mock[ConsignmentRepository]
+    when(consignmentRepoMock.consignmentHasFiles(any[UUID])).thenReturn(mockResponse)
+
+    val consignmentService: ConsignmentService = new ConsignmentService(consignmentRepoMock, FixedTimeSource, fixedUuidSource)
+    val response: Boolean = consignmentService.consignmentHasFiles(UUID.randomUUID()).await()
+    verify(consignmentRepoMock, times(1)).consignmentHasFiles(any[UUID])
```
suzannehamilton, line 102: This check is unnecessary - as a rule of thumb, you only need to verify that a mock is called if it has side-effects (such as modifying data, or sending an email or something).

In this case, what we really care about is that given the test setup, the response is correct, which you're checking for below.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -88,4 +88,34 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
...
     response should be(None)
   }
+
+  "consignmentHasFiles" should "return true when files already associated with provided consignment id" in {
+    val fixedUuidSource = new FixedUUIDSource()
+    val fileUuid1 = UUID.randomUUID()
+    val fileUuid2 = UUID.randomUUID()
+    val mockResponse: Future[Boolean] = Future.successful(true)
+    val consignmentRepoMock = mock[ConsignmentRepository]
+    when(consignmentRepoMock.consignmentHasFiles(any[UUID])).thenReturn(mockResponse)
```
suzannehamilton, line 98: Lots of random UUIDs in this test. Is there any reason this can't check for the specific UUID rather than `any[UUID]`? It would make the test less likely to pass for the wrong reason, and it makes it easier to read too, because you can tell which of the three UUIDs are actually important.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -88,4 +88,34 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
 
     response should be(None)
   }
+
+  "consignmentHasFiles" should "return true when files already associated with provided consignment id" in {
+    val fixedUuidSource = new FixedUUIDSource()
```
suzannehamilton, line 93: Do we need to use FixedUUIDSource and UUID.randomUUID in the same test?

To be honest, I'd recommend using fixed UUIDs in unit tests. That way we don't have to worry about the state of FixedUUIDSource, or random test failures from a random data source*. It also means that if you see a UUID in the output of a failed test, you can track it back to a specific bit of test data.

\* Random test data is really useful in some circumstances, but you have to be able to re-run your tests with the same random seed, which we haven't set up here.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -88,4 +88,34 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
...
+    when(consignmentRepoMock.consignmentHasFiles(any[UUID])).thenReturn(mockResponse)
+
+    val consignmentService: ConsignmentService = new ConsignmentService(consignmentRepoMock, FixedTimeSource, fixedUuidSource)
+    val response: Boolean = consignmentService.consignmentHasFiles(UUID.randomUUID()).await()
+    verify(consignmentRepoMock, times(1)).consignmentHasFiles(any[UUID])
+
+    response should be(true)
+  }
+
+  "consignmentHasFiles" should "return false when no files associated with provided consignment id" in {
```
suzannehamilton, line 107: Same suggestions as above for this test.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
TomJKing: The ConsignmentService constructor requires a UUID source which the FixedUUIDSource gives

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
TomJKing: Removed the random UUID

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationAuthoriser.scala
TomJKing: Reverted back to authorisation tag type, and implemented separate sangria middleware to check consignment state validation as discussed

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationTag.scala
TomJKing: Moved to graphql package as separately discussed

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
TomJKing: check removed

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -88,4 +88,28 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
 
     response should be(None)
   }
+
+  "consignmentHasFiles" should "return true when files already associated with provided consignment id" in {
+    val fixedUuidSource = new FixedUUIDSource()
+    val mockResponse: Future[Boolean] = Future.successful(true)
+    val consignmentRepoMock = mock[ConsignmentRepository]
+    when(consignmentRepoMock.consignmentHasFiles(fixedFileUUID)).thenReturn(mockResponse)
```
suzannehamilton, line 96: Should `fixedFileUUID` by `consignmentId`? `consignmentHasFiles` takes a consignment ID rather than a file ID.

src/main/scala/uk/gov/nationalarchives/tdr/api/consignmentstatevalidation/ConsignmentStateTags.scala
```diff
@@ -0,0 +1,28 @@
...
+import sangria.execution.BeforeFieldResult
+import sangria.schema.{Argument, Context}
+import uk.gov.nationalarchives.tdr.api.graphql.{ConsignmentApiContext, ValidationTag}
+import uk.gov.nationalarchives.tdr.api.graphql.validation.UserOwnsConsignment
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ConsignmentStateTag extends ValidationTag
+
+case class ValidateNoPreviousUploadForConsignment[T](argument: Argument[T]) extends ConsignmentStateTag {
```
suzannehamilton, line 14: Sorry I didn't spot this before, but is this tag only going to be used on one endpoint? If so, you don't need the `arg match` statement below - you can just pull the argument out of the context like the simpler auth tags do:

```
val fileInput = ctx.arg[FileInputArg]("addFilesInput")
```

If we need to need to reuse this tag on another endpoint in future, we can think about commonising it then, because `UserOwnsConsignment` sounds quite auth-specific and might not be relevant here. (And if it _is_ relevant, we should rename it, or possibly think of a nicer way of wrapping the consignment ID than pattern matching every possible argument type.)

src/main/scala/uk/gov/nationalarchives/tdr/api/consignmentstatevalidation/ConsignmentStateValidator.scala
```diff
@@ -0,0 +1,43 @@
...
+      case v: ConsignmentStateTag => {
+        val validationResult = v.validate(ctx)
+
+        // Awaiting a Future is risky because the thread will block until the response is returned or the timeout is reached.
+        // It could cause the API to be slow because akka-http cannot assign threads to new requests while this one is
+        // blocked.
+        //
+        // We are only using Await because Sangria middleware does not support Futures like the main resolvers do. We should
+        // remove it when we find a way to do authorisation in a completely async way in Sangria.
+        Await.result(validationResult, 5 seconds)
```
suzannehamilton, line 34: This risky `Await.result` is now duplicated. It would be good to isolate it in one place, so that if we ever [fix it](https://national-archives.atlassian.net/browse/TDR-52), it will be fixed everywhere.

What do you think about moving the `Await` to the `validate` method of `Validation` tag? So it would look something like:

```
trait ValidationTag extends FieldTag {
  
  def validate(ctx: Context[ConsignmentApiContext, _])
              (implicit executionContext: ExecutionContext): BeforeFieldResult[ConsignmentApiContext, Unit] = {
    val validationResult = validate(ctx)

    // Comment about why this is bad
    Await.result(validationResult, 5 seconds)
  }

  def validateAsync(ctx: Context[ConsignmentApiContext, _])
              (implicit executionContext: ExecutionContext): Future[BeforeFieldResult[ConsignmentApiContext, Unit]]

  val continue: BeforeFieldResult[ConsignmentApiContext, Unit] = BeforeFieldResult(())
}
```

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
TomJKing: Grr, yes good point. Fixed

src/main/scala/uk/gov/nationalarchives/tdr/api/consignmentstatevalidation/ConsignmentStateTags.scala
TomJKing: Good point. Fixed

src/main/scala/uk/gov/nationalarchives/tdr/api/consignmentstatevalidation/ConsignmentStateValidator.scala
TomJKing: Implemented this so validate function is in one place.

PR #44, opened 2020-04-30
Split config into environment-specific files

suzannehamilton:
This will make it easier to add different CORS configuration to integration vs staging and production.

It will also make it easier to set up a development environment, because the local database connection config is in the dev-specific config file.

https://national-archives.atlassian.net/browse/TDR-113

### DIFF
--- a/src/main/resources/application.base.conf
+++ b/src/main/resources/application.base.conf
@@ -1,0 +1,22 @@
+auth = {
+  url = ${AUTH_URL}
+}
+
+frontend = {
+  url = ${FRONTEND_URL}
+}
+
+source = {
+  uuid = uk.gov.nationalarchives.tdr.api.service.RandomUUID
+}
+
+consignmentapi = {
+  profile = "slick.jdbc.PostgresProfile$"
+  db {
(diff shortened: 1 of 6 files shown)


src/main/resources/application.base.conf
```diff
@@ -0,0 +1,25 @@
...
+  url = ${AUTH_URL}
+}
+frontend = {
+  url = ${FRONTEND_URL}
+}
+
+source = {
+  uuid = uk.gov.nationalarchives.tdr.api.service.RandomUUID
+}
+
```
TomJKing, line 11: Extra line break

suzannehamilton: Fixed.

suzannehamilton:
The Readme's still up-to-date, because we didn't get around to adding the extra environment variables to it and now they're not necessary. Eventual consistency. 🙂

suzannehamilton:
I'm not going to merge this yet because https://github.com/nationalarchives/tdr-terraform-environments/pull/57 needs to be merged and deployed first.

PR #45, opened 2020-05-01
Tdr 131 add av endpoint

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -91,14 +90,14 @@ case class ValidateUserOwnsConsignment[T](argument: Argument[T]) extends Authori
   }
 }
 
-object ValidateUserOwnsFiles extends AuthorisationTag {
+case class ValidateUserOwnsFiles[T](argument: Argument[T]) extends AuthorisationTag {
```
suzannehamilton, line 93: In this case I think you can use `Argument[UserOwnsFile]` and drop the type parameter. The type param is only necessary in `ValidateUserOwnsConsignment` because it takes a UUID _or_ a custom type.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/AVMetadataRepository.scala
```diff
@@ -0,0 +1,15 @@
+package uk.gov.nationalarchives.tdr.api.db.repository
+
+import slick.jdbc.PostgresProfile.api._
+import uk.gov.nationalarchives.Tables.{AvmetadataRow, Avmetadata}
+
+import scala.concurrent.Future
+
+class AVMetadataRepository(db: Database) {
```
suzannehamilton, line 8: Minor point, but could you name these classes and methods `Antivirus...` rather than `AV...`? It makes it easier for someone who's not as familiar with the codebase to work out what's going on. I always read "AV" as "audiovisual" before I remember it means "antivirus" and I bet I'm not the only one.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
suzannehamilton: I wrote this before I realised we should be checking for a machine client token, so this might be a moot point.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/AVMetadataRepository.scala
TomJKing: Renamed from "av" to "antivirus"

suzannehamilton:
Let me know when this is ready for re-review.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -114,3 +114,16 @@ object ValidateUserOwnsFiles extends AuthorisationTag {
...
+
+object ValidateHasAntiVirusMetadataAccess extends SyncAuthorisationTag {
+  override def validateSync(ctx: Context[ConsignmentApiContext, _]): BeforeFieldResult[ConsignmentApiContext, Unit] = {
+    val token = ctx.ctx.accessToken
+    val antivirusAccess = token.backendChecksRoles.contains("anti_virus")
+
+    if (antivirusAccess) {
+      continue
+    } else {
+      throw AuthorisationException("No access to anti-virus metadata")
```
suzannehamilton, line 126: Just a thought: is there any useful info in the token, e.g. a user ID or something, that we could include in the logs to work out what client was trying to access the metadata? It would probably have to be something that's in a regular user token and a service client token.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -114,3 +114,16 @@ object ValidateUserOwnsFiles extends AuthorisationTag {
       })
   }
 }
+
+object ValidateHasAntiVirusMetadataAccess extends SyncAuthorisationTag {
+  override def validateSync(ctx: Context[ConsignmentApiContext, _]): BeforeFieldResult[ConsignmentApiContext, Unit] = {
+    val token = ctx.ctx.accessToken
+    val antivirusAccess = token.backendChecksRoles.contains("anti_virus")
```
suzannehamilton, line 121: There are going to be a few of these values, so it would be good for them to be constants somewhere so we're not checking against hard-coded strings.

Also, really minor thing, but we're writing "antivirus" as a single word everywhere else, so why is it "anti_virus" in the token?

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
```diff
@@ -47,6 +47,15 @@ object TestUtils  {
       .withClaim("user_id", userId)
       .build)
   )
+  def validBackendChecksToken(): OAuth2BearerToken = OAuth2BearerToken(tdrMock.getAccessToken(
```
suzannehamilton, line 50: Nitpick: this helper refers to "valid backend checks" but it's specifically for antivirus. I'd be inclined to pass the role as a parameter to make the test setup more explicit.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
TomJKing: Because thats how  I named it in Keycloak. Have changed it

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
TomJKing: Added the user id for the Keycloak generated service account user to the message. This should allow tracing back to the client etc.

PR #46, opened 2020-05-01
Allow cross-origin requests from localhost to integration

suzannehamilton:
Configure frontend domains as a list rather than as a single value.

Use this to allow cross-origin requests to the integration API from the integration frontend AND the local dev environment.

This makes it possible to do local frontend development without setting up the rest of the TDR stack.

I haven't fully tested this locally because of the [current issues with local dev](https://national-archives.atlassian.net/browse/TDR-111). I have checked that the MVC site can still make requests to the API. If there are any issues, they'll be with JavaScript requests, but that might be easier to spot when it's deployed to integration.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/http/ApiServer.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/http/ApiServer.scala
@@ -3,8 +3,8 @@ package uk.gov.nationalarchives.tdr.api.http
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.Materializer
+import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.Logger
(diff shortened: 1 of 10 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/http/ApiServer.scala
```diff
@@ -21,7 +21,9 @@ object ApiServer extends App {
 
   scala.sys.addShutdownHook(() -> shutdown())
 
-  Http().bindAndHandle(route, "0.0.0.0", PORT)
+  val routes = new Routes(ConfigFactory.load())
```
suzannehamilton, line 24: These changes are because I needed to inject the config object in order to test different config values. I think it's a useful change in general, though, because we're now only loading the config once rather than every time it's called.

TomJKing: Good idea

PR #49, opened 2020-05-22
Add more information to ValidateUserOwnsConsignment error message

suzannehamilton:
This makes it easier to debug issues when requests with this auth tag fail, for example if the Keycloak client is misconfigured and doesn't include the user ID.


suzannehamilton:
Good point. 🤦‍♀️  Now fixed!

PR #53, opened 2020-06-19
Split up endpoints into graphql and admin

MancunianSam:
This isn't going to be merged but I've raised this so people can see the diff a bit easier. 
It's part of the spike to see about splitting the graphql server into two separate endpoints.


suzannehamilton:
Good to see that this doesn't require any changes to the field definitions. 👍

PR #55, opened 2020-06-25
Tdr 330 ffid api mutation

MancunianSam:


build.sbt
```diff
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-optics" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
-  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.22",
+  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.24",
```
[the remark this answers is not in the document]
MancunianSam: Yeah, that's the latest one in S3 in 24. I think the snapshot versions are the unreleased versions if that makes sense so the next release will be 0.0.25

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FFIDMetadataFields.scala
```diff
@@ -0,0 +1,45 @@
...
+                               containerSignatureFileVersion: String,
+                               method: String,
+                               extension: Option[String] = None,
+                               identificationBasis: String,
+                               puid: Option[String],
+                               datetime: Long)
+    implicit val AddFFFIDMetadataInputType: InputObjectType[FFIDMetadataInput] = deriveInputObjectType[FFIDMetadataInput]()
+    implicit val FFIDMetadataType: ObjectType[Unit, FFIDMetadata] = deriveObjectType[Unit, FFIDMetadata]()
+
+    val FileMetadataInputArg = Argument("addFFIDMetadataInput", AddFFFIDMetadataInputType)
```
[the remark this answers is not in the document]
MancunianSam: Yep, makes sense, I've renamed it.

src/main/resources/application.conf
```diff
@@ -1,7 +1,7 @@
 include "application.base"
 
 auth = {
-  url = "http://localhost:8081/auth"
+  url = "https://auth.tdr-integration.nationalarchives.gov.uk/auth"
```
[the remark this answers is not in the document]
MancunianSam: No, I didn't mean to commit that. I've set it back.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FFIDMetadataFields.scala
```diff
@@ -0,0 +1,45 @@
...
+  case class FFIDMetadata(fileId: UUID,
+                          software: String,
+                          softwareVersion: String,
+                          binarySignatureFileVersion: String,
+                          containerSignatureFileVersion: String,
+                          method: String,
+                          extension: Option[String] = None,
+                          identificationBasis: String,
+                          puid: Option[String],
+                          datetime: Long)
```
suzannehamilton, line 23: Is this the timestamp that we think is for auditing? If so, should it be generated by the API so that the API is responsible for all the integrity data, rather than trusting the client application?

MancunianSam: Normally yes but the reason I did it like this is that @Dclipsham said that we would need the scan date from the siegfried result so that's why I've passed it in.

build.sbt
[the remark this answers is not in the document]
TomJKing: Gotcha

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FFIDMetadataFields.scala
MancunianSam: I've changed it now so the API generates the datetime.

suzannehamilton: Yeah, I'd have thought the date we're interested in is when the data is saved, since that's when it becomes available to other parts of the system. Do you know if that's good enough for David's purposes, or is there some other reason to save the scan date?

Just to check - will the Siegfried version information that we're saving tell us what version of the file format signatures is being used? Assuming that's the important thing if we need to debug any unexpected results?

MancunianSam: Yep, that's what the binarySignatureFileVersion and containerSignatureFileVersion do. There's a conversation on slack recently where I've asked what he needs. And for the scan date, I can't see any reason why the few minutes it takes to process through the system is going to make much difference unless people are uploading at 2 minutes to midnight which seems unlikely.

PR #56, opened 2020-06-30
Tdr 367 get file metadata

MancunianSam:


src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -145,3 +146,17 @@ object ValidateHasChecksumMetadataAccess extends SyncAuthorisationTag {
...
+object ValidateHasFFIDMetadataAccess extends SyncAuthorisationTag {
+  override def validateSync(ctx: Context[ConsignmentApiContext, _]): BeforeFieldResult[ConsignmentApiContext, Unit] = {
+    val token = ctx.ctx.accessToken
+    val fileFormatAccess = token.backendChecksRoles.contains(fileFormatRole)
+
+    if (fileFormatAccess) {
+      continue
+    } else {
+      val tokenUserId = token.userId.getOrElse("")
+      throw AuthorisationException(s"User '$tokenUserId' does not have permission to update file format metadata")
```
suzannehamilton, line 159: The error message says "does not have permission to update the metadata", but it's being used in a query field which isn't updating anything. Should we have a separate permission about fetching file data?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ClientFileMetadataService.scala
```diff
@@ -1,17 +1,26 @@
...
 
 import scala.concurrent.{ExecutionContext, Future}
 
 class ClientFileMetadataService(clientFileMetadataRepository: ClientFileMetadataRepository, uuidSource: UUIDSource)
                                (implicit val executionContext: ExecutionContext) {
 
+  def getClientFileMetadata(fileId: UUID): Future[ClientFileMetadata] = {
+    clientFileMetadataRepository.getClientFileMetadata(fileId).map(_.head).map(rowToOutput).recover {
+      case _: NoSuchElementException => throw InputDataException(s"Could not find metadata for file $fileId")
+      case e: SQLException => throw InputDataException(e.getLocalizedMessage)
```
suzannehamilton, line 20: Can you add the original exceptions as causes of the InputDataException so we don't lose the original stack trace?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ClientFileMetadataRouteSpec.scala
```diff
@@ -117,6 +153,23 @@ class ClientFileMetadataRouteSpec extends AnyFlatSpec with Matchers with TestReq
     ps.executeUpdate()
   }
 
+  // scalastyle:off magic.number
```
suzannehamilton, line 156: Should this be re-enabled after the function, or does that happen automatically?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ClientFileMetadataService.scala
MancunianSam: So the current case class looks like this :
`case class InputDataException(message: String) extends Exception(message)`
Do you mean change it to this and leave it at that?
`case class InputDataException(message: String, cause: Throwable) extends Exception(message, cause)`

The problem with this is that there is already a line in FileMetadataService which throws an exception without an original exception.
`case None => throw InputDataException(s"The property does not exist")` 
In this case, the property that has been passed in is wrong so this is very much an InputDataException.

And in terms of keeping the original stack trace, do we really need it here? The InputDataException will give the line number which threw it and from there we can easily work out what's going on without 100 lines of slick classes in the stack trace.

I'm open to adding it but at the moment it causes issues with the way it's being used elsewhere and doesn't seem to give us much benefit.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
MancunianSam: Yeah good point. I've added in a `client_file_metadata` role to control access.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ClientFileMetadataRouteSpec.scala
MancunianSam: No it needs re-enabling, I've added it back in.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ClientFileMetadataService.scala
suzannehamilton: If this was Java, you'd have an overloaded constructor. In Scala, I guess the best way to do it is with a default value:

```
class InputDataException(message: String, cause: Throwable = null) extends Exception(message, cause)
```

I'd always include the original exception to avoid throwing away information. It's hard to predict in advance what info will be useful when you're debugging something, and it's so easy to include an exception cause that my rule of thumb is to just do it.

MancunianSam: Fair point. I've changed it to be an option in the case class because null seems to be bad in Scala (https://docs.scala-lang.org/overviews/scala-book/no-null-values.html) although I have had to use .orNull because it's a java class.

PR #57, opened 2020-07-09
TDR-337-Add file avmetadata endpoint

Lholbourn:
These changes create functions to check the progress of AVMetadata for files in the AVMetadata repository. This will facilitate creating a progress page for TDR users to see how many of their files have been processed and how long it may be until they can add metadata for their consignment. 

At the moment, this deals only with AVMetadata progress and not progress of other file checks, thus the object used has the class properties of two Ints. This will need to be changed when we deal with more data. I think a Map may work well as we can set the type of data as the key and have a List/Array/Seq of the progress numbers as the value. I haven't made this change yet as I would like to know if anyone has other opinions on this. 

This change will need to be accompanied by a change to the [tdr-generated-graphql](https://github.com/nationalarchives/tdr-generated-graphql) repo before the endpoint can be used by other projects.


src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
```diff
@@ -105,4 +111,13 @@ object TestUtils  {
     ps.setTimestamp(4, Timestamp.from(FixedTimeSource.now))
     ps.executeUpdate()
   }
+
+  def addAntivirusMetadata(fileId: String, consignmentId: String): Unit = {
+    val sql = s"insert into AVMetadata (FileId, Result, Datetime) VALUES (?, ?, ?)"
+    val ps: PreparedStatement = DbConnection.db.source.createConnection().prepareStatement(sql)
+    ps.setString(1, s"${fileId}")
+    ps.setString(2, consignmentId)
```
suzannehamilton, line 119: It looks like the second parameter of this INSERT statement should be the antivirus result, rather than the consignment ID.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
```diff
@@ -1,6 +1,7 @@
 package uk.gov.nationalarchives.tdr.api.utils
 
 import java.sql.{PreparedStatement, Timestamp}
+import java.time.Instant
```
suzannehamilton, line 4: Looks like there might be some unused imports to clear up. It's a shame ScalaStyle doesn't catch this!

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -13,8 +13,11 @@ import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
 object ConsignmentFields {
   case class Consignment(consignmentid: Option[UUID] = None, userid: UUID, seriesid: UUID)
   case class AddConsignmentInput(seriesid: UUID)
+  case class ConsignmentFileMetadataProgress(processedFiles: Int, totalFiles: Int)
```
suzannehamilton, line 16: Now I'm looking at this in the context of the other GraphQL queries, I'm wondering if we should actually integrate this into the Consignment endpoint rather than creating a new one. 🤔 

Do you want to pair on this again? There are some Sangria tricks we can use, and it's probably easier to go through them in person. It shouldn't affect the rest of your code - it's just about this GraphQL layer.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
```diff
@@ -0,0 +1,110 @@
...
+    val consignmentId = "c6f78fef-704a-46a8-82c0-afa465199e65"
+    val fileOneId = "20e0676a-f0a1-4051-9540-e7df1344ac10"
+    val fileTwoId = "b5111f11-4dca-4f92-8239-505da567b9df"
+    var count = 0
+
+    TestUtils.createConsignment(UUID.fromString(consignmentId), userId)
+    TestUtils.createFile(UUID.fromString(fileOneId), UUID.fromString(consignmentId))
+    TestUtils.createFile(UUID.fromString(fileTwoId), UUID.fromString(consignmentId))
+
+    while (count < 7) {
```
suzannehamilton, line 99: A slightly more idiomatic way of doing this in Scala would be:

```
(1 to 7).foreach { _ => TestUtils.addAntivirusMetadata(fileOneId, consignmentId) }
```

(I hope my syntax is right there - I haven't checked it.)

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
Lholbourn: Ah you're right! Will change :ok_hand:

Lholbourn:
> The logic for this looks good. I've had a thought about the GraphQL endpoint which it would be good to pair on if you've got some time today?
> 
> We should also write some integration tests (the route specs), but that will be easier once we've figured out the interface.

That would be great yeah! I'm free all day :+1:

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
Lholbourn: Yeah I wasn't 100% on how to make it work well with Scala, I'll change that!

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
Lholbourn: I think that could be a really good idea yeah! :raised_hands:

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -5,16 +5,41 @@ import java.util.UUID
...
+  implicit private val ConsignmentType: ObjectType[Unit, Consignment] = ObjectType(
+    "Consignment",
+    fields[Unit, Consignment](
+      Field("consignmentid", OptionType(UuidType), resolve = _.value.consignmentid),
+      Field("userid", UuidType, resolve = _.value.userid),
+      Field("seriesid", UuidType, resolve = _.value.seriesid),
+      Field(
+        "totalFiles",
+        IntType,
+        resolve = context => DeferTotalFiles(context.value.consignmentid)
```
suzannehamilton, line 33: This didn't occur to me on Friday, but we could use the GraphQL query argument, which is a `UUID`, rather than the result, which is an `Option[UUID]` so we don't have all the awkward Option handling in DeferTotalFiles and DeferFileProgress.

I think you'd replace `context.value.consignmentid` with `context.args.arg(ConsignmentIdArg)`.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -5,16 +5,41 @@ import java.util.UUID
...
+      ),
+      Field(
+        "fileCheckProgress",
```
suzannehamilton, line 36: The nested structure has made me think we should actually put more of the file check results into this object. So something like:

```
fileChecks {
  antivirus {
    filesProcessed
  },
  fileFormatId {
    filesProcessed
  }
}
```

Which would let us add the results into each section as well as the number of files processed. What do you think, and how much work would it be to shuffle the result classes around to fit that pattern?

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
```diff
@@ -1,15 +1,29 @@
...
   def addFiles(fileRows: Seq[FileRow]): Future[Seq[Tables.FileRow]] = {
     db.run(insertQuery ++= fileRows)
   }
+
+  def countFilesInConsignment(consignmentId: UUID): Future[Int] = {
+    val query = File.filter(_.consignmentid === consignmentId).length
+    db.run(query.result)
+  }
+
+  def countProcessedFilesInConsignment(consignmentId: UUID): Future[Int] = {
```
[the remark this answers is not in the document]
Lholbourn: Sure will do!

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
Lholbourn: I think that's a good way forward, I was picturing something similar when I started the nested stuff. I don't think it would be too difficult to add in the other processes once we're at that point. I've made the classes a bit more general rather than all about AV and the AV object rests inside of the fileChecks object so I think all we need to do is change that case class to include the objects, not many other big changes I can think of though!

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
```diff
@@ -1,15 +1,29 @@
...
+    val query = (for {
```
suzannehamilton, line 24: While I was debugging the test failure yesterday, I noticed that you can remove the `for` comprehension to simplify this query a bit:

```
val query = Avmetadata.join(File)
      .on(_.fileid === _.fileid)
      .filter(_._2.consignmentid === consignmentId)
      .groupBy(_._1.fileid)
      .map(_._1)
      .length
    db.run(query.result)
```

It now reads a bit more like the SQL query it's based on.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
suzannehamilton: We tried this yesterday and it didn't work - it looks like `context.args` contains the arguments for the _field_, not the arguments for the parent query.

It would be good to make `context.value.consignmentid` a non-optional value, but it's a change to the existing API so we'd have to be careful not to break any of the clients.

suzannehamilton:
Looks like there are still a couple of open comments about the structure of the fields and the names of the resolvers. I'll do another review once those are updated.

PR #60, opened 2020-07-29
Update logging configuration

techncl:
Configure API to log everything at info level and above to STDOUT.
This fixes the issue with almost nothing being logged to Cloudwatch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-optics" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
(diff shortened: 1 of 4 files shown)


suzannehamilton:
@TomJKing Would you mind taking a look at this one? Nigel and I paired on it.

build.sbt
```diff
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-optics" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
-  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.25",
+  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.26",
```
TomJKing, line 45: There is a version 0.0.27 available for the `consignment-api-db`. Not sure you want to use this version instead?

PR #67, opened 2020-08-13
Tdr 402 log responses

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
```diff
@@ -23,6 +25,29 @@ class Routes(val config: Config) extends Cors {
...
+  def isNon200Status(status: StatusCode): Boolean = {
+    val code = status.intValue.toString
+    !code.startsWith("2")
+  }
+
+  def logging: Directive0 = {
+    def logNon200Response(req: HttpRequest)(res: RouteResult): Unit = {
+      res match {
+        case Complete(resp) =>
+          if (isNon200Status(resp.status)) {
```
suzannehamilton, line 41: What about `resp.status.isFailure`? I think that covers all the cases we're interested in, and it means you don't have to inspect the status code number.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
```diff
@@ -23,6 +25,29 @@ class Routes(val config: Config) extends Cors {
...
+            val responseEntity = unmarshalEntity(resp.entity)
+            logger.info(s"Non 200 Response: $req\nStatus Code: ${resp.status}\nResponse Entity: $responseEntity")
```
suzannehamilton, line 43: Any idea how the new lines get rendered in CloudWatch? CW has an annoying tendency to split up multiline error messages - I'm hoping there's some way of configuring the logging to avoid that, but I haven't dug into the details.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
```diff
@@ -23,6 +25,29 @@ class Routes(val config: Config) extends Cors {
   val ttlSeconds: Int = 10
```
suzannehamilton, line 25: I just noticed the Logger name is wrong on line 24. ☝️ 

`Logger(classOf[Routes])` is [probably the thing to use](https://github.com/lightbend/scala-logging#using-scala-logging), because it includes the package name as well as the class name.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
```diff
@@ -23,6 +25,29 @@ class Routes(val config: Config) extends Cors {
...
+  def logging: Directive0 = {
+    def logNon200Response(req: HttpRequest)(res: RouteResult): Unit = {
+      res match {
+        case Complete(resp) =>
+          if (isNon200Status(resp.status)) {
+            val responseEntity = unmarshalEntity(resp.entity)
+            logger.info(s"Non 200 Response: $req\nStatus Code: ${resp.status}\nResponse Entity: $responseEntity")
+          }
+        case Rejected(reason) =>
+          logger.info(s"Rejected Reason: " + reason.mkString(", "))
```
suzannehamilton, line 46: Have you got any thoughts about the level we use for these messages? Given that we control the clients, maybe this should be at least a warning? Genuinely not sure.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
suzannehamilton, line 43: What do you think about logging the whole request body, which is probably going to be the key to debugging failed requests?

TomJKing: No idea, but removed the line breaks.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
TomJKing: Feels like Warning makes the most sense to me, so changed to warning.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
TomJKing: Whole request body is logged.

suzannehamilton: It sounds like making CloudWatch handle multline logs is possible but a bit fiddly. I've created [TDR-468](https://national-archives.atlassian.net/browse/TDR-468) so that we can come back to this, since it affects all the backend apps.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/RouteLogging.scala
```diff
@@ -0,0 +1,32 @@
...
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.server.RouteResult.{Complete, Rejected}
+import akka.http.scaladsl.server.directives.{DebuggingDirectives, LoggingMagnet}
+import akka.http.scaladsl.server.{Directive0, RouteResult}
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.Materializer
+import com.typesafe.scalalogging.Logger
+
+object RouteLogging {
+  implicit val actorSystem: ActorSystem = ActorSystem("graphql-server")
```
suzannehamilton, line 13: Is this line necessary? I wouldn't have expected this object to be using a completely different actor system to Routes.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/RouteLogging.scala
```diff
@@ -0,0 +1,32 @@
...
+  implicit val materializer: Materializer = Materializer(actorSystem)
+
+  private val logger = Logger(classOf[Routes])
```
suzannehamilton, line 16: Should this be `classOf[RouteLogging]` since you've moved this code into its own object?

src/main/resources/logback.xml
```diff
@@ -11,4 +11,7 @@
     <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
+    <root level="warn">
```
suzannehamilton, line 14: Is this necessary? The level defined on line 11 should define the _minimum_ level to log, so you shouldn't need to explicitly add higher log levels.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/RouteLoggingSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  "200 response" should "not be logged as warning" in {
+    val resp: HttpResponse = createHttpResponse(StatusCodes.OK)
+
+    Post("/") ~> RouteLogging.logging(complete(resp))
+
+    logCaptor.getWarnLogs.isEmpty shouldBe(true)
+  }
+
+  "non 200 response" should "be logged as warning" in {
+    val expectedWarningLogEntry = "Non 200 Response - Request: HttpRequest(HttpMethod(POST),http://example.com/," +
```
suzannehamilton, line 31: Checking exact log messages is normally very fragile - you don't want the test to fail if the message changes very slightly. You could check that the message contains the important parts of this, like request body and the HTTP status.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/RouteLoggingSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+      "List(),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1)), Request Entity: FulfilledFuture(), " +
```
suzannehamilton, line 32: Is the request entity `FulfilledFuture` correct? Does it correctly log the body if it's a regular GraphQL request?

src/main/scala/uk/gov/nationalarchives/tdr/api/http/RouteLogging.scala
TomJKing: Removed logging as an abstracted object

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -25,21 +27,25 @@ import scala.util.{Failure, Success}
 
 object GraphQLServer {
 
+  private val logger = Logger(classOf[GraphQLServer])
+
+  private def handleException(marshaller: ResultMarshaller, errorCode: String, message: String): HandledException = {
+    val node = marshaller.scalarNode(errorCode, "String", Set.empty)
+    val additionalFields = Map("code" -> node)
+    logger.warn(s"$message, Codes: ${additionalFields}")
```
suzannehamilton, line 35: Minor thing, but I'd just include the `errorCode` instead of `additionalFields`. The latter is just the structure we need to put it in for Sangria to add it to the GraphQL response - the error code is the important thing to have in the logs.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -25,21 +27,25 @@ import scala.util.{Failure, Success}
 
 object GraphQLServer {
 
+  private val logger = Logger(classOf[GraphQLServer])
+
+  private def handleException(marshaller: ResultMarshaller, errorCode: String, message: String): HandledException = {
```
suzannehamilton, line 32: This looks like a sensible place to put the GraphQL logging. 👍

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -110,3 +115,5 @@ object GraphQLServer {
       }
   }
 }
+
+class GraphQLServer
```
suzannehamilton, line 119: This is new - is it needed?

suzannehamilton (commented):
Looking good, thanks. A couple of really small comments.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
TomJKing: If we want to use the `classOf` for the logger as GraphQLServer is an object. Otherwise will need to hard code the value for the Logger: `Logger("...GraphQLServer")`

TomJKing: Was just getting the `classOf` for the logger to work. Removed it

suzannehamilton (approved):
Looks good, thanks!

PR #69, opened 2020-08-25
Tdr 158 use body code

MancunianSam:


TomJKing (approved):
Looks good. Will need to update Keycloak to add the code, instead of the name value to the body attribute of the department groups.

PR #75, opened 2020-09-10
Tdr 412 add ffidmetadata

Lholbourn:
This will allow us to show user the progress of FFIDMetadata processing in the front-end project. The changes are made in line with those made previously with anti-virus and checksum metadata. Includes tests for both the repository method and the consignment service. 

This was also tested manually with a local version of the API and Postman and behaves as expected.


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -27,8 +27,6 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
     conn.close()
   }
 
-  //"fileChecks": { "antivirusProgress": { "filesProcessed":  1},  "checksumProgress": {"filesProcessed": 2}}
```
suzannehamilton, line 30: Thanks for tidying that up.

suzannehamilton (commented):
This looks good to me!

There's just one thing about the way the FFID Metadata is saved that's made me realise we need to check something with Sam. I'll message you both to organise a quick chat.

suzannehamilton (approved):
Good to go. 🚀 

As discussed with Sam, it looks a bit weird that none of our test examples add data to the `FFIDMetadataMatches` table, but that is OK when all we care about is the file format progress. We'll decide whether to worry about missing rows from the matches table when we get to the failure path pages.

Lholbourn:
> Good to go. 
> 
> As discussed with Sam, it looks a bit weird that none of our test examples add data to the `FFIDMetadataMatches` table, but that is OK when all we care about is the file format progress. We'll decide whether to worry about missing rows from the matches table when we get to the failure path pages.

Awesome, I'll keep it in mind for future!

PR #76, opened 2020-09-21
Tdr 206 validate checksum

MancunianSam:
The tests won't pass on this branch yet because the update to the consignment-api-db repo isn't merged and deployed but the rest of the code is ready to look at.


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
```diff
@@ -26,4 +26,12 @@ class FileMetadataRepository(db: Database) {
       .length
     db.run(query.result)
   }
+
+  def addChecksumValidationResult(fileId: UUID, validationResult: Option[Boolean]): Future[Int] = {
+    val query = for {
+      file <- File if file.fileid === fileId
+    } yield file.checksummatches
+    val update = query.update(validationResult)
+    db.run(update)
```
suzannehamilton, line 35: Is it possible to run this update in the same transaction as saving the metadata? That would avoid the situation where one update succeeds and the other fails, leaving them out of date.

I haven't tried this with Slick before, but it looks like you can add `transactionally` to the end of an action which contains updates to multiple tables: https://scala-slick.org/doc/3.2.0/dbio.html#transactions-and-pinned-sessions

MancunianSam: Yep. I've changed it so both the insert and update run in the same transaction and I've refactored it a bit so this only happens with a checksum update.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
```diff
@@ -78,18 +142,65 @@ class FileMetadataServiceSpec extends AnyFlatSpec with MockitoSugar with Matcher
...
+    service.addFileMetadata(AddFileMetadataInput("SHA256ServerSideChecksum", fileId, "checksum"), Some(UUID.randomUUID())).futureValue
+    verify(fileMetadataRepositoryMock).addChecksumMetadata(any[FilemetadataRow], any[Option[Boolean]])
+  }
+
+  "addFileMetadata" should "fail if the update is not for a checksum" in {
+    val fileMetadataRepositoryMock = mock[FileMetadataRepository]
+    val filePropertyRepositoryMock = mock[FilePropertyRepository]
+    val clientFileMetadataServiceMock = mock[ClientFileMetadataService]
+    val fileId = UUID.randomUUID()
+//    val propertyId = UUID.randomUUID()
```
suzannehamilton, line 172: Commented-out code. Left over from when you were fixing the tests?

MancunianSam: Ah yeah. I've removed it cheers.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -4,33 +4,40 @@ import java.sql.{SQLException, Timestamp}
...
+      case "SHA256ServerSideChecksum" =>
+        (for {
+          fileProperty <- getFileProperty(addFileMetadataInput.filePropertyName)
+          cfm <- clientFileMetadataService.getClientFileMetadata(addFileMetadataInput.fileId) if fileProperty.isDefined
+          row <- fileMetadataRepository.addChecksumMetadata(row(fileProperty.get), cfm.checksum.map(_ == addFileMetadataInput.value))
+        } yield FileMetadata(fileProperty.flatMap(_.name).get, row.fileid, row.value)) recover {
+          case e: Throwable => throw InputDataException(e.getLocalizedMessage, Some(e))
         }
-      case None => throw InputDataException(s"The property does not exist", Option.empty)
+      case _ => Future.failed(InputDataException("We are only expecting checksum updates for now"))
```
suzannehamilton, line 34: Can you stick the actual `filePropertyName` in the error message so we can work out what happened if we ever hit this error?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileMetadataRouteSpec.scala
```diff
@@ -100,6 +101,17 @@ class FileMetadataRouteSpec extends AnyFlatSpec with Matchers with TestRequest w
     checkNoFileMetadataAdded()
   }
 
+  "addFileMetadata" should "add the checksum validation result if this is a checksum update" in {
+    createClientFileMetadata(defaultFileId)
+    runTestMutation("mutation_alldata", validBackendChecksToken("checksum"))
+    checkValidationResultExists(defaultFileId)
+  }
+
+  "addFileMetadata" should "add the checksum validation result if this is not a checksum update" in {
```
suzannehamilton, line 110: Is there a "not" missing from the "add the checksum validation result"?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileMetadataRouteSpec.scala
```diff
@@ -111,7 +123,7 @@ class FileMetadataRouteSpec extends AnyFlatSpec with Matchers with TestRequest w
 
   private def createFileProperty = {
     val sql = "insert into FileProperty (PropertyId , Name, Description, Shortname) " +
-      "VALUES (?, 'ServerSideChecksum', 'The checksum calculated after upload', 'Checksum')"
+      "VALUES (?, 'SHA256ServerSideChecksum', 'The checksum calculated after upload', 'Checksum')"
```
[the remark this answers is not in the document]
MancunianSam: Yep good point, I've put it in the fields object, seems like the right place.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: Yep, makes sense, I've added it in.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileMetadataRouteSpec.scala
MancunianSam: The copy and paste demon strikes again. I've updated it, cheers.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
suzannehamilton: That's useful, thanks, but I meant adding `addFileMetadataInput.filePropertyName`. An error like "Unexpected property 'MD5Hash'" is easier to debug than just "Unexpected property" because you can start asking "Wait, what's trying to save MD5 hashes?" 😄

MancunianSam: That's what I've done I think https://github.com/nationalarchives/tdr-consignment-api/pull/76/files#diff-a2f15bd61b5d5ae2b2905333419fa072R36
It adds the property name in. Or is it the wording of the error you mean?

suzannehamilton: Ahh, I think I was somehow looking at an older version of the code. Sorry!

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -4,33 +4,42 @@ import java.sql.{SQLException, Timestamp}
...
+      case SHA256ServerSideChecksum =>
+        (for {
+          fileProperty <- getFileProperty(addFileMetadataInput.filePropertyName)
+          cfm <- clientFileMetadataService.getClientFileMetadata(addFileMetadataInput.fileId) if fileProperty.isDefined
+          row <- fileMetadataRepository.addChecksumMetadata(row(fileProperty.get), cfm.checksum.map(_ == addFileMetadataInput.value))
+        } yield FileMetadata(fileProperty.flatMap(_.name).get, row.fileid, row.value)) recover {
+          case e: Throwable => throw InputDataException(e.getLocalizedMessage, Some(e))
         }
-      case None => throw InputDataException(s"The property does not exist", Option.empty)
+      case _ => Future.failed(InputDataException(s"$filePropertyName found. We are only expecting checksum updates for now"))
```
suzannehamilton, line 36: Not worth it now (given that we'll probably have to reshuffle the metadata code either for the auditing or the custom metadata anyway), but it might be good to see whether Sangria plays nicely with [GraphQL enums](https://graphql.org/learn/schema/#enumeration-types). Maybe we won't need this error handling at all if we can reject invalid property names at the schema level?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
```diff
@@ -78,18 +143,55 @@ class FileMetadataServiceSpec extends AnyFlatSpec with MockitoSugar with Matcher
     result.value should equal(value)
   }
 
+  "addFileMetadata" should "add the checksum validation result if the update is for a checksum" in {
```
suzannehamilton, line 146: Does this test check anything that's not covered by the tests on lines 61 and 88? It seems like they're better tests because they check the specific value of the result boolean, not just that any boolean is saved. Unless I'm missing something?

suzannehamilton (commented):
Just one small comment on the tests, and one idea for the future.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
MancunianSam: No, you're right, the other two check the same argument on the same method but more specifically. I've removed this one.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: That would be good. Worth looking at after we've finished with the metadata code then.

suzannehamilton (approved):

suzannehamilton (approved):

PR #79, opened 2020-10-08
Tdr 62 remove created date

MancunianSam:


TomJKing (commented):
Looks fine. I noticed that there are some deprecation warnings FileServiceSpec: `adaptation of an empty argument list by inserting () is deprecated: this is unlikely to be what you want`

Looks like you just need to change: `val mockConsignmentResponse = Future.successful()` to `val mockConsignmentResponse = Future.successful(())`

MancunianSam:
Yep good point. I've changed it.

TomJKing (approved):

PR #80, opened 2020-10-16
Tdr 510 get parent name and files

Lholbourn:
This PR adds the ability to query the API and get the parent folder name for a consignment. It initially used a much more manual process but was refactored to be included in the normal `getConsignment` fields using the deferred resolver to keep new API endpoints to a minimum. It includes alterations to existing tests as well as new tests to cover the new `ConsignmentService`.

Once this change is merged, it will need to be accompanied by a change to the [generated GraphQL project](https://github.com/nationalarchives/tdr-generated-graphql).


src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -21,10 +21,23 @@ class ConsignmentRepositorySpec extends AnyFlatSpec with ScalaFutures with Match
...
-    consignmentRepository.addParentFolder(consignmentId, Option("TEST PARENT FOLDER NAME")).futureValue
+    consignmentRepository.addParentFolder(consignmentId, Option("TEST ADD PARENT FOLDER NAME")).futureValue
 
     val parentFolderName = consignmentRepository.getConsignment(consignmentId).futureValue.map(consignment => consignment.parentfolder)
 
-    parentFolderName should contain only Some("TEST PARENT FOLDER NAME")
+    parentFolderName should contain only Some("TEST ADD PARENT FOLDER NAME")
+  }
+
+  "getParentFolder" should "get parent folder name for a consignment" in {
```
suzannehamilton, line 31: It would also be good to have a test of the case where the name hasn't been saved yet, to show that the query doesn't throw an error.

suzannehamilton (commented):
Looks good, thanks! Just a test suggestion and a question about the GraphQL query nesting.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -17,6 +17,7 @@ object ConsignmentFields {
   case class ChecksumProgress(filesProcessed: Int)
   case class FFIDProgress(filesProcessed: Int)
   case class FileChecks(antivirusProgress: AntivirusProgress, checksumProgress: ChecksumProgress, ffidProgress: FFIDProgress)
+  case class ParentFolder(parentFolder: Option[String] = None)
```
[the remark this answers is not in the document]
Lholbourn: Oh yeah, I hadn't thought about that. I initially had it nested because I created a new API endpoint, then didn't think to undo that when I changed it to be part of the deferred resolver, etc. I'll try to un-nest so it's a less complex data structure :+1:

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
Lholbourn: I'll add that in! :+1:

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -42,6 +42,11 @@ object ConsignmentFields {
         "fileChecks",
         FileChecksType,
         resolve = context => DeferFileChecksProgress(context.value.consignmentid)
+      ),
+      Field(
+        "parentFolder",
+        StringType,
```
suzannehamilton, line 48: Should this be `OptionType(StringType)`, since the parent folder name might not exist yet?

Lholbourn: Ah yes you're right! :eyes:

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
```diff
@@ -17,10 +17,15 @@ class DeferredResolver extends sangria.execution.deferred.DeferredResolver[Consi
...
+          id => context.consignmentService.getConsignmentParentFolder(id)
+        ).getOrElse(Future.successful(0))
       case other => throw UnsupportedDeferError(other)
     }
   }
 }
 
 case class DeferTotalFiles(consignmentId: Option[UUID]) extends Deferred[Int]
 case class DeferFileChecksProgress(consignmentId: Option[UUID]) extends Deferred[FileChecks]
+case class DeferParentFolder(consignmentId: Option[UUID]) extends Deferred[String]
```
suzannehamilton, line 31: Now that the parent folder is an optional value, should this deferred resolver also return an optional type? Then you could get rid of the default on line 23.

Lholbourn: I think that makes much more sense, yeah :smile:

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
```diff
@@ -17,10 +17,15 @@ class DeferredResolver extends sangria.execution.deferred.DeferredResolver[Consi
         consignmentId.map(
           id => context.consignmentService.getConsignmentFileProgress(id)
         ).getOrElse(Future.successful(0))
+      case DeferParentFolder(consignmentId) =>
+        consignmentId.map(
+          id => context.consignmentService.getConsignmentParentFolder(id)
+        ).get
```
suzannehamilton, line 23: Is the `.get` necessary, now that this case returns an optional value?

Lholbourn: I did initially try without the `.get` but kept getting a type mismatch error:
Screenshot from 2020-10-22 09-38-48
But if I change the return type for the deferred resolver then the other case statements have type mismatches. There might be a better way to get around it though that I'm not thinking of!

suzannehamilton: Oh, that's interesting. I didn't realise it was `Option[Future[String]]`. Probably easiest to have a quick call to decide how to handle this - give me a shout on Slack.

Lholbourn: Sure will do! :+1:

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -47,6 +49,8 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
   val runTestMutation: (String, OAuth2BearerToken) => GraphqlMutationData = runTestRequest[GraphqlMutationData](addConsignmentJsonFilePrefix)
   val expectedQueryResponse: String => GraphqlQueryData = getDataFromFile[GraphqlQueryData](getConsignmentJsonFilePrefix)
   val expectedMutationResponse: String => GraphqlMutationData = getDataFromFile[GraphqlMutationData](addConsignmentJsonFilePrefix)
+  val runParentFolderTestQuery: (String, OAuth2BearerToken) => GraphqlQueryData = runTestRequest[GraphqlQueryData](getParentFolderAndFilesJsonFilesPrefix)
+  val expectedParentQueryResponse: String => GraphqlQueryData = getDataFromFile[GraphqlQueryData](getParentFolderAndFilesJsonFilesPrefix)
```
suzannehamilton, line 53: Are these two values used anywhere? Or are they not needed because you've added the test of the parent folder to the existing `getConsignment` test?

Lholbourn: Ah they're not used anywhere, forgot to remove them! :+1:

suzannehamilton (approved):

PR #82, opened 2020-11-04
Fields for export task

MancunianSam:
There are three new fields.
* getFiles which returns all files for a consignment
* updateExportLocation which is called by the export task when it has uploaded the tar to s3
* updateTransferInitiated which is called by the frontend when the transfer is initiated.


suzannehamilton (commented):
I still need to review the tests, sorry, but I'll post the questions I've added so far.

build.sbt
```diff
@@ -13,7 +13,8 @@ scalacOptions ++= Seq("-deprecation", "-feature")
 
 resolvers ++= Seq[Resolver](
   "Sonatype Releases" at "https://dl.bintray.com/mockito/maven/",
-  "TDR Releases" at "s3://tdr-releases-mgmt"
+  "TDR Releases" at "s3://tdr-releases-mgmt",
+  "TDR Snapshots" at "s3://tdr-snapshots-mgmt"
```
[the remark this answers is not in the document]
MancunianSam: It was to get the build to work while the consignment-api-data changes were in code review. I was going to leave it in but I don''t think it''s necessary. Now the database changes are deployed, I've changed the version and removed the snapshot resolver.

build.sbt
```diff
@@ -42,7 +43,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-optics" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
-  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.33",
+  "uk.gov.nationalarchives" %% "consignment-api-db" % "0.0.34-SNAPSHOT",
```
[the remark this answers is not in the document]
MancunianSam: Yep, I've changed it.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -20,6 +23,20 @@ class ConsignmentRepository(db: Database) {
...
   }
 
+  def updateExportLocation(exportLocationInput: ConsignmentFields.UpdateExportLocationInput, timestamp: Timestamp): Future[Int] = {
+    val update = Consignment.filter(_.consignmentid === exportLocationInput.consignmentId)
+      .map(c => (c.exportlocation, c.exportdatetime))
+      .update((Option(exportLocationInput.exportLocation), Option(timestamp)))
+    db.run(update)
+  }
+
+  def updateTransferInitiated(consignmentId: UUID, userId: Option[UUID], timestamp: Timestamp): Future[Int] = {
```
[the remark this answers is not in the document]
MancunianSam: The [authoriser](https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala#L31) just checks to see if it's a valid keycloak token which it is without the user id parameter. We could change the authoriser to make sure there is one but for now, we don't.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
```diff
@@ -11,6 +11,11 @@ import scala.concurrent.Future
 class FileRepository(db: Database) {
   private val insertQuery = File  returning File.map(_.fileid)into ((file, fileid) => file.copy(fileid = fileid))
 
+  def getFiles(consignmentId: UUID): Future[Seq[Tables.FileRow]] = {
```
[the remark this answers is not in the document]
MancunianSam: Yeah, most likely. Because it's not being called by the browser, it might be ok.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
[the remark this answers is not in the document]
suzannehamilton: Ah, that's interesting. Can you remember if the user ID is None when the token comes from a service client?

If service client tokens have a user ID, maybe we should reject it in the authoriser? If they _don't_ have a user ID, maybe we should reject requests where we have to keep an audit trail of the user?

src/test/resources/json/updateexportlocation_mutation_all.json
```diff
@@ -0,0 +1,9 @@
+{
+  "query": "mutation updateExportLocation($updateExportLocationInput:UpdateExportLocationInput!){updateExportLocation(exportLocation:$updateExportLocationInput)}",
+  "variables": {
+    "updateExportLocationInput": {
+      "consignmentId": "6e3b76c4-1745-4467-8ac5-b4dd736e1b3e",
+      "exportLocation": "6e3b76c4-1745-4467-8ac5-b4dd736e1b3e.tar.gz"
```
suzannehamilton, line 6: This has made me think: do you know whether the export location will be a full S3 URL, or a path within a bucket? I don't really mind either way - I guess it depends on how it will be used. If it's an S3 URL, should we validate the structure?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -153,15 +157,58 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
...
+
+  "updateTransferInitiated" should "update the transfer initiated date correctly" in {
+    insertConsignment()
+    val prefix = "json/updatetransferinitiated_"
+    val expectedResponse = getDataFromFile[GraphqlMutationTransferInitiated](prefix)("data_all")
+    val response: GraphqlMutationTransferInitiated = runTestRequest[GraphqlMutationTransferInitiated](prefix)("mutation_all", validUserToken())
+    response.data should equal(expectedResponse.data)
+    val field = getConsignmentField(UUID.fromString("6e3b76c4-1745-4467-8ac5-b4dd736e1b3e"), _)
+    Option(field("TransferInitiatedDatetime")).isDefined should equal(true)
+    field("TransferInitiatedBy") should equal("4ab14990-ed63-4615-8336-56fbb9960300")
```
suzannehamilton, line 178: Really minor thing, but could this be `userId` rather than the hard-coded UUID so it's obvious where it comes from?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -153,15 +157,58 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
...
+  }
+
+  private def insertConsignment(): Unit = {
```
suzannehamilton, line 181: Does `TestUtils.createConsignment` do the same thing?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -102,6 +107,27 @@ class FileRouteSpec extends AnyFlatSpec with Matchers with TestRequest with Befo
...
+    val fileIdTwo = "0f70f657-8b19-4ab6-9813-33a8223fec84"
+    insertFile(consignmentId, fileIdOne)
+    insertFile(consignmentId, fileIdTwo)
+    val expectedResponse = expectedQueryResponse("data_all")
+    val response = runTestQuery("mutation_alldata", validBackendChecksToken("export"))
+
+    expectedResponse.data.get.getFiles should equal(response.data.get.getFiles)
+  }
+
+  private def insertFile(consignmentId: String, fileId: String): Unit = {
```
suzannehamilton, line 122: Could reuse `TestUtils.createFile`?

src/test/resources/json/updateexportlocation_mutation_all.json
MancunianSam: The way I've written the consignment export at the moment it's a path within the bucket. I think it's fine for now. We know which bucket it's going to be in. If it turns out to be a problem later, we can always change it.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
MancunianSam: Yep, makes sense, I've changed it.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
MancunianSam: It does indeed, cheers. I've changed it to use that.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
MancunianSam: Yep, I've changed it to use that.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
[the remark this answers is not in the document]
MancunianSam: I've changed the auth-utils `userId` method so it returns a `UUID` instead of `Option[String]` The method which gets the token will return a failure if the userId isn't set so this is fine.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -162,15 +166,46 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
...
+
+  private def checkConsignmentExists(consignmentId: UUID): Unit = {
+    val rs = getConsignment(consignmentId)
     rs.next()
     rs.getString("ConsignmentId") should equal(consignmentId.toString)
   }
 
+  private def getConsignmentField(consignmentId: UUID, field: String): String = {
+    val rs = getConsignment(consignmentId)
+    rs.next()
```
suzannehamilton, line 205: Is it possible to move the `rs.next()` into the method that executes the query? It's a bit confusing when it's out of context here.

Also, I'd name the variable `result` or something - `rs` is also quite opaque when it's away from the context of a `ResultSet`.

suzannehamilton (commented):
Just one tiny suggestion to make the tests easier to understand.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
MancunianSam: Yep, fair point, I've changed it.

suzannehamilton (approved):

PR #83, opened 2020-11-06
Make ID non-optional in the Consignment domain model

suzannehamilton:
Refactor the Consignment model so that the ID is compulsory. It's not clear why this was an Option in the first place, but removing it makes the code simpler because the deferred resolvers don't need to decide what to do if the value is None.


suzannehamilton:
@techncl This is the refactoring we talked about the other day. I thought I'd try it out while I was waiting for something to build, and it was simpler than I expected!

techncl (approved):
Hi Suzanne! This looks a lot cleaner and seems fine when I test it :+1:

PR #84, opened 2020-11-09
Add series and department to api query

techncl:


src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -210,4 +209,90 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers
...
+    val userId = UUID.fromString("ceb8831c-2017-11eb-adc1-0242ac120002")
+    val datetime = Timestamp.from(FixedTimeSource.now)
+    val parentFolder = Option("Parent_folder")
+    val bodyId = UUID.fromString("8eae8ed8-201c-11eb-adc1-0242ac120002")
+    val seriesName = Option("Mock series")
+    val seriesCode = Option("Mock series")
+    val seriesDescription = Option("Series description")
+    val bodyName = Option("Mock department")
+    val bodyCode = Option("Mock department")
+    val bodyDescription = Option("Body description")
```
suzannehamilton, line 238: What do you think about extracting out some methods which create the consignment, series and body, to remove some of the lengthy setup from the test itself? They could provide default values for the things that we don't care about in this test, such as the body description or the consignment creation time.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
```diff
@@ -16,8 +16,9 @@ class FileRepositorySpec extends AnyFlatSpec with ScalaFutures with Matchers {
     val db = DbConnection.db
     val fileRepository = new FileRepository(db)
     val consignmentId = UUID.fromString("c03fd4be-58c1-4cee-8d3c-d162bb4f7c02")
+    val seriesId = UUID.fromString("1436ad43-73a2-4489-a774-85fa95daff32")
 
-    TestUtils.createConsignment(consignmentId, userId)
+    TestUtils.createConsignment(consignmentId, seriesId, userId)
```
suzannehamilton, line 21: One way to avoid having to update all of these existing tests is to give the `seriesId` parameter a default value in the `createConsignment` method. Then the tests in `FileRepositorySpec` could stay the same, and the new tests could pass a specific value.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -17,6 +17,7 @@ object ConsignmentFields {
   case class ChecksumProgress(filesProcessed: Int)
   case class FFIDProgress(filesProcessed: Int)
   case class FileChecks(antivirusProgress: AntivirusProgress, checksumProgress: ChecksumProgress, ffidProgress: FFIDProgress)
+  case class TransferringBody(bodyid: UUID, name: Option[String], code: Option[String], description: Option[String])
```
suzannehamilton, line 20: Do we just need the name at the moment? I'd remove the other fields for now. We know there are some database changes coming, and it will be easier to apply those if the API doesn't refer to fields which might be removed or renamed.

src/test/resources/json/getconsignment_query_alldata.json
```diff
@@ -1 +1 @@
-{"query":"{getConsignment(consignmentid: \"b130e097-2edc-4e67-a7e9-5364a09ae9cb\"){seriesid, consignmentid, userid, totalFiles, fileChecks {antivirusProgress {filesProcessed}, checksumProgress {filesProcessed}, ffidProgress {filesProcessed}}, parentFolder}}"}
\ No newline at end of file
+{"query":"{getConsignment(consignmentid: \"b130e097-2edc-4e67-a7e9-5364a09ae9cb\") {seriesid, consignmentid, userid, totalFiles, fileChecks {antivirusProgress {filesProcessed}, checksumProgress {filesProcessed}, ffidProgress {filesProcessed}}, parentFolder, series {name}, transferringBody {name}}}"}
```
suzannehamilton, line 1: Does the frontend use the code rather than the name in the series drop-down? If so, I think it's the code that we should be depending on here.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -29,6 +30,18 @@ class ConsignmentService(
     consignments.map(rows => rows.headOption.map(row => Consignment(row.consignmentid, row.userid, row.seriesid)))
   }
 
+  def getSeriesOfConsignment(consignmentId: UUID): Future[Option[Series]] = {
+    val consignment: Future[Seq[ConsignmentResult]] = consignmentRepository.getSeriesAndBodyOfConsignment(consignmentId)
+    consignment.map(rows => rows.headOption.map(
+      row => Series(row.series.seriesid, row.series.bodyid, row.series.name, row.series.code, row.series.description)))
+  }
+
+  def getTbOfConsignment(consignmentId: UUID): Future[Option[TransferringBody]] = {
```
suzannehamilton, line 39: Could `Tb` be `TransferringBody`? It's more verbose, but I don't think we use TB as an acronym often enough for these to be obvious.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -25,13 +23,22 @@ class ConsignmentRepository(db: Database) {
     db.run(query.result)
   }
 
+  def getSeriesAndBodyOfConsignment(consignmentId: UUID)(implicit executionContext: ExecutionContext): Future[Seq[ConsignmentResult]] = {
```
suzannehamilton, line 26: Now that we're using separate resolvers for the transferring body and the series, I think it's worth splitting this up into two separate repository methods.

It means more code, but I think it's less confusing for future developers because they won't have to work out why the series resolver gets the transferring body details and then throws them away.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -25,13 +23,22 @@ class ConsignmentRepository(db: Database) {
...
+    val query = Consignment.join(Series)
+      .on(_.seriesid === _.seriesid).join(Body)
+      .on(_._2.bodyid === _.bodyid)
+      .filter(_._1._1.consignmentid === consignmentId)
+      .map(rows => (rows._1._1, rows._1._2, rows._2))
+    db.run(query.result).map(rows => rows.map(result => ConsignmentResult(result._1, result._2, result._3)))
```
suzannehamilton, line 32: Does the consignment need to be returned at all? I don't think it's being used in the service.

suzannehamilton (commented):
A few suggestions for the tests and splitting up the repository method. Otherwise, this looks good. 👍

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
techncl: Ah yes, that is a good idea; I completely forgot you could do this in Scala :smile:

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
techncl: Ok, sure!

src/test/resources/json/getconsignment_query_alldata.json
techncl: Ah, yes, you are correct it's the code; I confused it with the Body

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
techncl: Sure!

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
techncl: No, it doesn't; I'll remove it. I thought since it's being returned by the database, it might as well be returned from the method but I guess if it comes in handy in the future, someone can add it back, they would only need to change the services that called it and that would probably be a small thing

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
techncl: Ok, sure!

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
techncl: Ah yes, this is a good idea; it was something I was thinking about too.

suzannehamilton:
Looks like you've got a failing test in Jenkins.

techncl:
> Looks like you've got a failing test in Jenkins.

Yes, weirdly, it was because the series had the "name" field instead of the "code", which I could have sworn I changed twice; I think it might have reverted the changes somehow.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -1,213 +1,145 @@
...
 
-class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with ScalaFutures {
+class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMocksAfterEachTest with Matchers with ScalaFutures {
   implicit val executionContext: ExecutionContext = ExecutionContext.Implicits.global
-  private val fixedConsignmentUUID = UUID.fromString("6e3b76c4-1745-4467-8ac5-b4dd736e1b3e")
+
+  val fixedUuidSource: FixedUUIDSource = mock[FixedUUIDSource]
+  val bodyId: UUID = UUID.fromString("8eae8ed8-201c-11eb-adc1-0242ac120002")
+  val userId: UUID = UUID.randomUUID()
+  val seriesId: UUID = UUID.randomUUID()
```
suzannehamilton, line 26: Minor thing, but can you generate a couple of UUIDs and use them for the userId and seriesId, like we do for the bodyId?

It makes the tests easier to debug, because if you see "expected 'foo' but got 'bar'", you can look up where the value came from.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -1,213 +1,145 @@
...
-
+    val parentFolderResult: Option[String] = consignmentService.getConsignmentParentFolder(consignmentId).futureValue
     parentFolderResult shouldBe parentFolder
   }
+
+  "getSeriesOfConsignment" should "return the series for a given consignment" in {
+    val mockSeries = Seq(SeriesRow(seriesId, bodyId, seriesName, seriesCode, seriesDescription))
+    when(consignmentRepoMock.getSeriesOfConsignment(consignmentId)).thenReturn(Future.successful(mockSeries))
+
+    val expectedSeries: SeriesFields.Series = consignmentService.getSeriesOfConsignment(consignmentId).futureValue.get
```
suzannehamilton, line 130: Variable naming: this could just be `series`. It's not the expected series, it's the actual one that the service returns.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -1,213 +1,145 @@
...
+    expectedSeries.name shouldBe mockSeries.head.name
+    expectedSeries.code shouldBe mockSeries.head.code
+    expectedSeries.description shouldBe mockSeries.head.description
+  }
+
+  "getTransferringBodyOfConsignment" should "return the transferring body for a given consignment" in {
+    val mockBody = Seq(BodyRow(bodyId, bodyName, bodyCode, bodyDescription))
+    when(consignmentRepoMock.getTransferringBodyOfConsignment(consignmentId)).thenReturn(Future.successful(mockBody))
+
+    val expectedBody: ConsignmentFields.TransferringBody = consignmentService.getTransferringBodyOfConsignment(consignmentId).futureValue.get
```
suzannehamilton, line 142: Ditto for "expected" in this name.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
suzannehamilton: Is this change still to be done?

suzannehamilton (commented):
Looks good! Just a couple of small naming comments, and (I think) one open comment from before.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
techncl: Ah, yes; good point!

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
techncl: Ohhh, I see; I thought "expected" meant what you expect to receive from the service.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
techncl: Sure!

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -1,213 +1,145 @@
...
 import uk.gov.nationalarchives.tdr.api.utils.{FixedTimeSource, FixedUUIDSource}
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with ScalaFutures {
+class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMocksAfterEachTest with Matchers with ScalaFutures {
   implicit val executionContext: ExecutionContext = ExecutionContext.Implicits.global
-  private val fixedConsignmentUUID = UUID.fromString("6e3b76c4-1745-4467-8ac5-b4dd736e1b3e")
+
+  val fixedUuidSource: FixedUUIDSource = mock[FixedUUIDSource]
```
suzannehamilton, line 23: Sorry, just noticed this. Didn't we use the UUIDSource trait yesterday rather than the FixedUUIDSource? If not, that's my mistake. 🙂 

The UUIDSource trait is what the service depends on, so that's what we should mock. The FixedUUIDSource is an alternative implementation for the integration tests, which this test doesn't need because it uses mocks rather than fake implementations.

suzannehamilton (commented):
Looks good! Just spotted one last thing.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
techncl: Ah, I see. Cool, I'll change it; thanks.

suzannehamilton (approved):

techncl:
:+1:

PR #85, opened 2020-11-18
Use apiuser instead of root user

TomJKing:


suzannehamilton:
@MancunianSam Could you take a look at this one? Does it address the main issue?

MancunianSam:
Yep, I think this solves our problem. Just one thing which I didn't notice before, could you change the name of the folder from `play` to something else as the consignment api isn't a play app.

MancunianSam (approved):
LGTM

PR #97, opened 2020-12-04
Refactor saving parent folder

Lholbourn:
These changes refactor the functions saving the parent folder name to the consignment table. Now that the front end is reliably saving the parent folder, it can be a `String` rather than `Option[String]`. I also amended some tests to reflect this change.


suzannehamilton (approved):
Looks good to me!

PR #103, opened 2020-12-15
Move client file metadata to FileMetadata table

MancunianSam:

PR #107, opened 2021-01-04
Tdr 716 save transfer agreement consignment

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-consignment-api-data/pull/68


src/main/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementService.scala
```diff
@@ -1,48 +1,96 @@
...
-    TransferAgreement(row.consignmentid,
-      row.allpublicrecords,
-      row.allcrowncopyright,
-      row.allenglish,
-      row.appraisalselectionsignedoff,
-      row.sensitivityreviewsignedoff,
-      Some(row.transferagreementid),
-      isAgreementComplete(row)
-    )
+  private def agreed(field: Option[Boolean]): Boolean = field.isDefined && field.get
```
suzannehamilton, line 71: You can do `field.getOrElse(false)`, which is a bit neater because you don't have to use `.get` which looks unsafe even though it's OK here because of the `.isDefined`.

Hopefully this is a moot point once we've done [TDR-733](https://national-archives.atlassian.net/browse/TDR-733) and made all the fields in this table non-nullable.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementService.scala
```diff
@@ -1,48 +1,96 @@
...
+
+  private def isAgreementComplete(propertyNameToValue: Map[String, Option[Boolean]]): Boolean = {
+    transferAgreementProperties.map(p => {
+      propertyNameToValue(p)
```
suzannehamilton, line 75: Probably safer to do a `.getOrElse` here too, so that if a field is missing then it returns `false` rather than throwing a NoSuchElementException because it's not in the map.

src/test/resources/scripts/init.sql
```diff
@@ -28,15 +28,22 @@ CREATE TABLE IF NOT EXISTS Consignment (
...
+  Description varchar(255),
+  Shortname varchar(255),
+  PRIMARY KEY (PropertyId)
+);
+
+CREATE TABLE IF NOT EXISTS ConsignmentMetadata (
+    MetadataId uuid,
+    ConsignmentId uuid not null,
+    PropertyId uuid not null,
+    Value varchar(255) not null,
```
suzannehamilton, line 43: Should these be nullable for now, to match the real DB?

suzannehamilton (commented):
A few suggestions about how the repo and and service are structured. What do you think?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementService.scala
```diff
@@ -1,48 +1,96 @@
...
                               (implicit val executionContext: ExecutionContext) {
 
-  def addTransferAgreement(input: AddTransferAgreementInput): Future[TransferAgreement] = {
-    val transferAgreementRow = TransferagreementRow(uuidSource.uuid, input.consignmentId,
-      input.allPublicRecords,
-      input.allCrownCopyright,
-      input.allEnglish,
-      input.appraisalSelectionSignedOff,
-      input.sensitivityReviewSignedOff)
+  implicit class StringUtils(value: Option[String]) {
```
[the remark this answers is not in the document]
TomJKing: Removed

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServer.scala
```diff
@@ -69,6 +69,7 @@ object GraphQLServer {
     }
   }
 
+  //noinspection ScalaStyle
```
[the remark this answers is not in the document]
TomJKing: Left in by mistake

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentMetadataRepository.scala
```diff
@@ -0,0 +1,50 @@
...
+      .filter(_._1.consignmentid === consignmentId)
+      .filter(_._2.name inSet propertyName.toSet)
+      .map(f => f._2.name.get -> f._1)
+    db.run(query.result).map(_.map(f => sqlRowToConsignmentMetadata(f._1, f._2)))
+  }
+}
+
+object ConsignmentMetadataRepository {
+
+  case class ConsignmentMetadataRowWithName(propertyName: String,
```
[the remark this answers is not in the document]
TomJKing: Further to the comments in https://github.com/nationalarchives/tdr-consignment-api/pull/103 to place this logic in the service layer would require something along  the lines of this approach:

**Inserting Transfer Agreement (TA) Input Values to ConsignmentMetadata DB Table**
1. retrieve the relevant TA properties from `ConsignmentProperty` DB table
2. create mapping between TA property name <- input value
3. generate the `ConsignmentmetadataRows` for insertion by iterating over the retrieved TA properties and adding the correct input value/ TA property id using the property name <- input value mapping

**Converting ConsignmentmetataRows to Transfer Agreement (TA)**
1. retrieve the relevant TA properties from `ConsignmentProperty` DB table
2. create mapping between TA property name <- TA property id
3. iterate over the TA properties and create a mapping between the TA input value to the TA property name using the mapping between TA property name <- TA property id
4. insert the correct input value to the `TransferAgreement` constructor using the map between the TA input value and the TA property name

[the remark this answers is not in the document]
TomJKing: Following discussion, database to be changed so that `PropertyId` is replaced by `PropertyName` in the `ConsignmentMetadata` table

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentMetadataRepositorySpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    val db = DbConnection.db
+    val consignmentMetadataRepository = new ConsignmentMetadataRepository(db)
+    val propertyId = UUID.randomUUID()
+    val consignmentId = UUID.fromString("d4c053c5-f83a-4547-aefe-878d496bc5d2")
+    addConsignmentProperty("ConsignmentProperty")
+    createConsignment(consignmentId, userId)
+    val input = Seq(ConsignmentmetadataRow(
+      UUID.randomUUID(), Some(consignmentId), Some("ConsignmentProperty"), Some("value"), Timestamp.from(Instant.now()), UUID.randomUUID()))
+    val result = consignmentMetadataRepository.addConsignmentMetadata(input).futureValue.head
+    result.propertyname should equal(Some("ConsignmentProperty"))
```
suzannehamilton, line 29: Is this testing the right thing? The test says "should add consignment metadata", but it only checks the return values. Does it need to check that the data is saved to the database?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
```diff
@@ -23,18 +24,19 @@ class TransfersAgreementRouteSpec extends AnyFlatSpec with Matchers with TestReq
...
   case class TransferAgreement(
-                                consignmentid: Option[UUID] = None,
+                                consignmentId: Option[UUID] = None,
                                 allPublicRecords: Option[Boolean] = None,
                                 allCrownCopyright: Option[Boolean] = None,
                                 allEnglish: Option[Boolean] = None,
                                 appraisalSelectionSignedOff: Option[Boolean] = None,
-                                sensitivityReviewSignedOff: Option[Boolean] = None,
-                                transferAgreementId: Option[UUID] = None
+                                initialOpenRecords: Option[Boolean] = None,
```
suzannehamilton, line 32: I agree with the casing changes, but are they backwards-compatible? Does this imply that the API client will need to be updated too?

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
```diff
@@ -197,5 +201,59 @@ object TestUtils {
 
     ps.executeUpdate()
   }
+
+  def addConsignmentProperty(name: String): Unit = {
+    // name is primary key check exists before attempting insert to table
+    if (!propertyExists(name)) {
```
suzannehamilton, line 207: I wonder if it would be better for init.sql to just have a list of standard property names to save before the tests are run?

suzannehamilton: If you want to stick with the current approach you can also add `ON CONFLICT(name) DO NOTHING`, which saves you the extra `SELECT` statement: https://stackoverflow.com/a/31742830/679924

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
```diff
@@ -197,5 +201,59 @@ object TestUtils {
...
+    ps.setString(2, consignmentId)
+    ps.setString(3, propertyName)
+    ps.setString(4, "Result of ConsignmentMetadata processing")
+    ps.setTimestamp(5, Timestamp.from(FixedTimeSource.now))
+    ps.setString(6, userId.toString)
+
+    ps.executeUpdate()
+  }
+
+  def createConsignmentMetadata(consignmentId: UUID): Unit = {
```
suzannehamilton, line 237: Should this be called `addTransferAgreementMetadata`? Otherwise it's hard to tell from the method name that it does transfer agreement stuff, and it also looks confusing similar to `addConsignmentMetadata`.

suzannehamilton (commented):
Looks good! Just a few comments on the tests.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
TomJKing: The casing changes now match the what is in the TransferAgreementFields.TransferAgreement case class.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
TomJKing: Looks like its an H2 Db we are using for the testing. So not sure the `DO NOTHING` statement is available.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
TomJKing: Fixed

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentMetadataRepositorySpec.scala
TomJKing: Added check that metatadata row has been added correctly

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
suzannehamilton: Then how did these tests pass before? I thought this class was used for deserializing the response in the test. 🤔

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestUtils.scala
suzannehamilton: OK, let's leave it for now.

Like we said on the call yesterday, these tests are getting a bit fragile because of the overlap in test data setup. But this is something that we need to think about more holistically.

suzannehamilton (approved):

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/TransfersAgreementRouteSpec.scala
TomJKing: Not sure, but if I revert the back to `consignmentid` the test fail. Curious

PR #112, opened 2021-01-21
Add static metadata

MancunianSam:
- Add new case class for the new static metadata.
- Add static metadata when the files are added.
- Add file metadata repository to file service object
- Update existing tests and add new ones.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
@@ -36,10 +36,17 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
 }
 
 object FileMetadataService {
+  case class StaticMetadata(name: String, value: String)
   val SHA256ClientSideChecksum = "SHA256ClientSideChecksum"
   val ClientSideOriginalFilepath = "ClientSideOriginalFilepath"
   val ClientSideFileLastModifiedDate = "ClientSideFileLastModifiedDate"
   val ClientSideFileSize = "ClientSideFileSize"
+  val RightsCopyright: StaticMetadata = StaticMetadata("RightsCopyright", "Crown Copyright")
+  val LegalStatus: StaticMetadata = StaticMetadata("LegalStatus", "Public Record")
+  val HeldBy: StaticMetadata = StaticMetadata("HeldBy", "TNA")
+  val Language: StaticMetadata = StaticMetadata("Language", "English")
+  val FoiExemptionCode: StaticMetadata = StaticMetadata("FoiExemptionCode", "open")
 
   val clientSideProperties = List(SHA256ClientSideChecksum, ClientSideOriginalFilepath, ClientSideFileLastModifiedDate, ClientSideFileSize)
+  val staticMetadataProperties = List(RightsCopyright, LegalStatus, HeldBy, Language, FoiExemptionCode)
 }
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
@@ -41,7 +41,13 @@ object FileMetadataService {
   val ClientSideOriginalFilepath = "ClientSideOriginalFilepath"
   val ClientSideFileLastModifiedDate = "ClientSideFileLastModifiedDate"
   val ClientSideFileSize = "ClientSideFileSize"
-  //There was a business decision made in the data enhancements workshop on 03/11/20 to store this metadata with these static values
+
+  /**
+   * Save default values for these properties because TDR currently only supports records which are Open, in English, etc.
+   * Users agree to these conditions at a consignment level, so it's OK to save these as defaults for every file.
+   * They need to be saved so they can be included in the export package.
+   * The defaults may be removed in future once we let users upload a wider variety of records.
+   */
   val RightsCopyright: StaticMetadata = StaticMetadata("RightsCopyright", "Crown Copyright")
   val LegalStatus: StaticMetadata = StaticMetadata("LegalStatus", "Public Record")
   val HeldBy: StaticMetadata = StaticMetadata("HeldBy", "TNA")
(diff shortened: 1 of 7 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -36,10 +36,17 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
 }
 
 object FileMetadataService {
+  case class StaticMetadata(name: String, value: String)
   val SHA256ClientSideChecksum = "SHA256ClientSideChecksum"
   val ClientSideOriginalFilepath = "ClientSideOriginalFilepath"
   val ClientSideFileLastModifiedDate = "ClientSideFileLastModifiedDate"
   val ClientSideFileSize = "ClientSideFileSize"
+  val RightsCopyright: StaticMetadata = StaticMetadata("RightsCopyright", "Crown Copyright")
```
suzannehamilton, line 44: I think it's worth adding a code comment explaining the reason for these fields. It's a business decision, so it's not obvious from the code why they need to be hard-coded.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/FixedUUIDSource.scala
```diff
@@ -17,7 +17,16 @@ class FixedUUIDSource extends UUIDSource {
       UUID.fromString("0824d319-9c58-4b15-81eb-b71fb460fa36"),
       UUID.fromString("15eddae1-f319-4da1-b2da-cda3bd8f840a"),
       UUID.fromString("854b1c99-5998-4126-a209-6f1bcc38bf74")
-    )(idx)
+    )
+
+    // The static uuids above are used when we're testing the correct id is returned.
+    // Some methods call the uuid() method up to 20 times but we don't need to know what the uuid is as we're not testing for it.
+    // So we just return a random uuid.
+    if(idx >= uuids.size) {
```
suzannehamilton, line 25: I already find this class quite confusing to work with. I'm worried that adding randomness will make it even harder!

Could we split this up, so each test can decide whether to use a fixed UUID source or a random UUID source?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -96,6 +109,38 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
...
+
+    val captor: ArgumentCaptor[Seq[FilemetadataRow]] = ArgumentCaptor.forClass(classOf[Seq[FilemetadataRow]])
+    when(fileMetadataRepositoryMock.addFileMetadata(captor.capture())).thenReturn(mockFileMetadataResponse)
+
+    val fileService = new FileService(fileRepositoryMock, consignmentRepositoryMock, fileMetadataRepositoryMock, FixedTimeSource, fixedUuidSource)
+    fileService.addFile(AddFilesInput(consignmentId, 1, "Parent folder name"), uuid).futureValue
+
+    val metadataRows: Seq[Tables.FilemetadataRow] = captor.getValue
+    metadataRows.size should equal(staticMetadataProperties.size)
+    staticMetadataProperties.foreach(property => {
```
suzannehamilton, line 136: Could we check each value individually rather than iterating over the list?

It feels like the test is repeating the logic of the code, and it would be nice to have a clear test stating what the expected default values are.

suzannehamilton (commented):
Just a few small comments, thanks!

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/FixedUUIDSource.scala
MancunianSam: Yeah ok. makes sense. I've had to remove one of the assertions in the `createFile" should "create three files given correct arguments`  test because with the `RandomUUIDSource`, I can't tell which uuids are coming from the source but other than that, the tests work using a `RandomUUIDSource`

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: Yeah good point, I've changed it to explicitly check for each of the names and values.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/FixedUUIDSource.scala
MancunianSam: Ignore that last comment. The `FixedUUIDSource` class lets us get the same uuids out each time it's called which allows us to test that IDs are being created correctly among other things. 
Now we're creating all the `FileMetadata` rows, the `FileService` class now calls `UUIDSource.uuid` more than 8 times which is why I added this random section in. But as you say, this isn't very neat. I did put in a `RandomUUIDSource` and used that in `FileServiceSpec` and this worked ok but `FileRouteSpec` is still using `FixedUUIDSource` (This is set in the test application.conf) and it's quite hard to change it to use the new RandomUUIDSource.

Anyway, long story short, I've added some more UUIDs to `FixedUUIDSource` so it doesn't run out when trying to create the `FileMetadata`.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: I've put a comment in saying more or less what it says on the Jira ticket.

suzannehamilton: Heh, I don't think that saying "it was a business decision" helps the person reading it understand the reason. 🙂 

I meant more like:

> Save default values for these properties because TDR currently only supports records which are Open, in English, etc. Users agree to these conditions at a consignment level, so it's OK to safe these as defaults for every file. They need to be saved so they can be included in the export package. The defaults may be removed in future once we let users upload a wider variety of records.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/FixedUUIDSource.scala
suzannehamilton: That's better for now, thanks. 👍 

I'd quite like to revisit FixedUUIDSource at some point because I think it makes it hard to work out which UUID is going to be used by which test, but I realise it's hard because of the way it's injected. I'll wait til I'm doing more work with these tests to try to see if I can come up with an option that's easier to understand.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: That's a lot better thanks :+1:

suzannehamilton (approved):

PR #114, opened 2021-01-25
Clean up test data after every test

suzannehamilton:
This fixes an issue where some tests would fail if only a subset of the tests were run, because of inter-dependencies. Now you can run all of the tests, or just one folder, or just one individual test.

This is quite a large change because every time I fixed a test, a different one would fail. It took a while to unpick all of the interdependencies.

Part of the problem was that init.sql gets run at unexpected times. It looks like it just gets executed at the start of the test suite, but it actually gets run every time the code creates a new database connection! Hopefully that will be a moot point once we can [move to a more realistic test database](https://national-archives.atlassian.net/browse/TDR-150), but I've mitigated the problems in the meantime by moving all the static data creation to a new trait called `TestDatabase`.

Some tests still use quite a lot of shared test data, but they no longer rely on the database being left in a particular state by  a previous test, so this is hopefully quite a big improvement.

Any new RepoSpecs or RouteSpecs should inherit from the new `TestDatabase` trait.

I timed the full test suite before and after the change. It still takes around 2 minutes, so the extra database teardown doesn't significantly affect the performance.

### DIFF
--- a/src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestDatabase.scala
+++ b/src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestDatabase.scala
@@ -1,0 +1,60 @@
+package uk.gov.nationalarchives.tdr.api.utils
+
+import java.sql.Connection
+
+import org.scalatest.{BeforeAndAfterEach, Suite}
+import uk.gov.nationalarchives.tdr.api.db.DbConnection
+import uk.gov.nationalarchives.tdr.api.service.FileMetadataService.{clientSideProperties, staticMetadataProperties}
+import uk.gov.nationalarchives.tdr.api.service.TransferAgreementService.transferAgreementProperties
+import uk.gov.nationalarchives.tdr.api.utils.TestUtils.{addConsignmentProperty, addFileProperty}
+
+/**
+ * This trait should be mixed into specs which access the test database.
+ *
+ * It provides a test database connection and cleans up all of the test data after every test.
+ */
+trait TestDatabase extends BeforeAndAfterEach {
+  this: Suite =>
+
+  val databaseConnection: Connection = DbConnection.db.source.createConnection()
+
+  override def beforeEach(): Unit = {
+    databaseConnection.prepareStatement("DELETE FROM FileMetadata").execute()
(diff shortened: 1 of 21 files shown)


src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TestDatabase.scala
```diff
@@ -0,0 +1,53 @@
...
+    databaseConnection.prepareStatement("DELETE FROM FileProperty").execute()
+    databaseConnection.prepareStatement("DELETE FROM FFIDMetadataMatches").execute()
+    databaseConnection.prepareStatement("DELETE FROM FFIDMetadata").execute()
+    databaseConnection.prepareStatement("DELETE FROM AVMetadata").execute()
+    databaseConnection.prepareStatement("DELETE FROM File").execute()
+    databaseConnection.prepareStatement("DELETE FROM ConsignmentMetadata").execute()
+    databaseConnection.prepareStatement("DELETE FROM ConsignmentProperty").execute()
+    databaseConnection.prepareStatement("DELETE FROM Consignment").execute()
+    databaseConnection.prepareStatement("DELETE FROM Series").execute()
+    databaseConnection.prepareStatement("DELETE FROM Body").execute()
```
suzannehamilton, line 32: I wanted to do something nicer here by using `DbConnection.db.run(DBIO.seq())` to run multiple queries, but it returns a Future, which is quite hard to get working with `beforeEach`. Might be worth revisiting at some point.

TomJKing (commented):
Looks good. One minor overall comment, the .json files you changed,  are missing a line break at the end.

suzannehamilton:
@TomJKing I've added those missing empty lines. 👍

TomJKing (approved):

suzannehamilton:
@TomJKing Would you mind taking another look at the TestDatabase trait? #112 got merged earlier, and it meant I had to add some extra setup to fix the merge conflicts: https://github.com/nationalarchives/tdr-consignment-api/pull/114/files#diff-4f5623c519ecd3799c140ec4fd287cc0869ad31da1d1cf2547391619b04090f8R39

TomJKing (approved):

PR #115, opened 2021-01-26
Add fields to support consignment export

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -44,6 +51,9 @@ object ConsignmentFields {
       Field("consignmentid", OptionType(UuidType), resolve = _.value.consignmentid),
       Field("userid", UuidType, resolve = _.value.userid),
       Field("seriesid", UuidType, resolve = _.value.seriesid),
+      Field("dateTime", OptionType(LocalDateTimeType), resolve = _.value.dateTime.get.toLocalDateTime),
+      Field("transferInitiatedDatetime", OptionType(LocalDateTimeType), resolve = _.value.transferInitiatedDatetime.get.toLocalDateTime),
+      Field("exportDatetime", OptionType(LocalDateTimeType), resolve = _.value.exportDatetime.get.toLocalDateTime),
       Field(
```
MancunianSam, line 57: What happens here if the `transferInitiatedDatetime` or `exportDatetime` aren't set? It looks like the call to `.get` would fail. Would it be better to do something like `.map(_.toLocalDateTime)`

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -83,7 +93,12 @@ object ConsignmentFields {
     Field("getConsignment", OptionType(ConsignmentType),
       arguments = ConsignmentIdArg :: Nil,
       resolve = ctx => ctx.ctx.consignmentService.getConsignment(ctx.arg(ConsignmentIdArg)),
-      tags = List(ValidateUserOwnsConsignment(ConsignmentIdArg))
+      tags = List(ValidateUserHasAccessToConsignment(ConsignmentIdArg))
+    ),
+    Field("getConsignmentExport", OptionType(ConsignmentType),
+      arguments = ConsignmentIdArg :: Nil,
```
MancunianSam, line 99: Do we need another field here. The new `ValidateUserHasAccessToConsignment` checks for export access so couldn't we just use the existing `getConsignment` field?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -1,18 +1,25 @@
...
 import uk.gov.nationalarchives.tdr.api.graphql._
 import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
 
 object ConsignmentFields {
 
-  case class Consignment(consignmentid: UUID, userid: UUID, seriesid: UUID)
+  case class Consignment(consignmentid: UUID,
+                         userid: UUID,
+                         seriesid: UUID,
+                         dateTime: Option[Timestamp],
```
suzannehamilton, line 19: This type is `java.sql.Timestamp`. It doesn't seem right to be working with database-specific types at the Sangria level of the code, because this layer shouldn't know anything about the database. I'd have expected this to be a type from `java.time`, and for the Service classes to translate between SQL timestamps and regular date-times.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: There's an error thrown in by the query, and the value is populated with null: 

"errors": [
        {
            "message": "Internal server error",
            "path": [
                "getConsignment",
                "transferInitiatedDatetime"
            ],
            "locations": [
                {
                    "column": 9,
                    "line": 5
                }
            ]
        }

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -1,18 +1,25 @@
 package uk.gov.nationalarchives.tdr.api.graphql.fields
 
-import java.util.UUID
+import java.sql.Timestamp
+import java.util.{Date, UUID}
```
suzannehamilton, line 4: `java.util.Date` is pretty much deprecated now because it was difficult to work with. You [should always use classes in the `java.time` package](https://codeblog.jonskeet.uk/2017/04/23/all-about-java-util-date/) (which was introduced in Java 8) unless you're dealing with legacy code. Unless there's some reason we have to use it here? 😢

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
MancunianSam: That's not what we want though is it? We want it to return the consignment with the dates empty instead of an error.

TomJKing: Will use map instead

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: Can do. Just means this will pull back all the additional fields needed for the export, but not necessarily used for anything else.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: Good point

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: Redundant import

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
MancunianSam: But the existing `getConsignment` query won't be requesting those fields and you can do a new query in the `tdr-generated-graphql` project like
```
query getConsignmentForExport($consignmentId: UUID!) {
    getConsignment(consignmentid: $consignmentId) {
        userid
        seriesid,
        datetime,
        transferInitiatedDatetime,
        exportDatetime      
    }
}
```
That keeps things separate without needing the new field.

TomJKing: Yeah, realised that after I wrote that comment!

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FieldTypes.scala
```diff
@@ -10,12 +11,30 @@ import scala.util.{Failure, Success, Try}
...
     case Failure(_) => Left(UuidCoercionViolation)
   }
 
+  private def parseDate(s: String): Either[ValueCoercionViolation, LocalDateTime] = Try(LocalDateTime.parse(s)) match {
+    case Success(localDateTime) => Right(localDateTime)
+    case Failure(_) => Left(LocalDateTimeCoercionViolation)
+  }
+
+  implicit val LocalDateTimeType: ScalarType[LocalDateTime] = ScalarType[LocalDateTime]("LocalDateTime",
+    coerceOutput = (ldt, _) => ldt.toString,
```
suzannehamilton, line 27: It looks like the output of `toString` depends on the precision of the input: https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#toString--

Is it possible to format it to a specific precision? Are the database timestamps in seconds? And can we include `Z` as the UTC time zone so there's no ambiguity in the API?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -65,4 +66,14 @@ class ConsignmentService(
   def getConsignmentParentFolder(consignmentId: UUID): Future[Option[String]] = {
     consignmentRepository.getParentFolder(consignmentId)
   }
+
+  private def convertRowToConsignment(row: ConsignmentRow): Consignment = {
+    Consignment(
+      row.consignmentid,
+      row.userid,
+      row.seriesid,
+      Some(row.datetime.toLocalDateTime),
```
suzannehamilton, line 75: Just wondering - if this value is non-nullable in the database, could it be non-optional in the Consignment model too?

src/test/resources/json/getconsignment_data_all.json
```diff
@@ -4,6 +4,9 @@
       "consignmentid":"b130e097-2edc-4e67-a7e9-5364a09ae9cb",
       "seriesid": "fde450c9-09aa-4ba8-b0df-13f9bac1e587",
       "userid":"4ab14990-ed63-4615-8336-56fbb9960300",
+      "dateTime": "2020-01-01T09:00",
```
suzannehamilton, line 7: Is this actually the consignment creation date? Could we give it a name like that? This field will be [renamed in the database](https://national-archives.atlassian.net/browse/TDR-640) at some point, but the API could use a more specific name already.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FieldTypes.scala
TomJKing: _Is it possible to format it to a specific precision?_ - Yes. What precision should be used, hours/seconds/milliseconds?
_Are the database timestamps in seconds?_ - DB timestamp precision is to milliseconds
_And can we include `Z` as the UTC time zone so there's no ambiguity in the API?_ - Not sure what you mean here, do you mean to include the 'UTC' timezone as part of the outputted Local Datetime string, ie: "2021-01-27 13:35:15 UTC"

suzannehamilton (approved):
🚀

PR #116, opened 2021-02-02
Fix flaky tests

suzannehamilton:
### Create foreign keys at same time as table in test database 

Merge the `ALTER TABLE` statements with the `CREATE TABLE` statements rather than keeping them separate. This means the foreign keys are only added if the tables don't already exist.

This solves a `ConcurrentModificationException` problem that was causing the tests to fail intermittently. init.sql is run every time a new database connection is opened, because it's specified as an init script in the H2 configuration in the tests' application.conf. If multiple connections were opened in parallel, for example because Sangria was resolving deferred resolvers in parallel, this sometimes meant that data was being read by one thread while another thread ran `ALTER TABLE` on the same table.

This change should mean that init.sql never modifies existing tables, which should fix the concurrency error.

### Fix flaky test by waiting for Future to finish

This test was failing intermittently because it was calling the repository method before waiting for the test setup to finish.

### DIFF
--- a/src/test/resources/scripts/init.sql
+++ b/src/test/resources/scripts/init.sql
@@ -78,17 +78,11 @@ CREATE TABLE IF NOT EXISTS FileMetadata (
     Value varchar(255) not null,
     Datetime timestamp not null DEFAULT CURRENT_TIMESTAMP,
     UserId uuid NOT NULL,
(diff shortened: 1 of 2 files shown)


MancunianSam (approved):
LGTM, cheers for sorting this out.

PR #117, opened 2021-02-03
Convert sql timestamp to ZonedDateTime

TomJKing:
To ensure clients can decode date times from API without custom decoders use ZonedDateTime for output

ZonedDateTime make the UTC timezone explicit to clients that need to use date times


MancunianSam (approved):
LGTM

PR #118, opened 2021-02-04
Add metadata field to consignment object

MancunianSam:
- Add a LocalDateTime scalar type
- Add service and repository methods to get metadata.
- Add service and repository tests
- Add deferred fields to consignment object.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
@@ -6,7 +6,7 @@ import java.util.UUID
 import sangria.execution.deferred.{Deferred, UnsupportedDeferError}
 import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields.{FileChecks, TransferringBody}
 import uk.gov.nationalarchives.tdr.api.graphql.fields.SeriesFields._
-import uk.gov.nationalarchives.tdr.api.service.FileMetadataService.FileMetadataValues
+import uk.gov.nationalarchives.tdr.api.service.FileMetadataService.{File, FileMetadataValues}
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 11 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -78,6 +81,11 @@ object ConsignmentFields {
         "transferringBody",
         OptionType(TransferringBodyType),
         resolve = context => DeferConsignmentBody(context.value.consignmentid)
+      ),
+      Field(
+        "fileMetadata",
```
suzannehamilton, line 86: Should this be called `files`, and each file has its own metadata? I'd have expected the GraphQL graph to be something like:

```
consignment {
  files {
    id
    metadata {
      legalStatus,
      sha256ClientSideChecksum
    }
  }
}
```

MancunianSam: Very good point. If we have a files object we can add other none metadata related stuff to the object as well. I've changed it.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
```diff
@@ -1,6 +1,7 @@
 package uk.gov.nationalarchives.tdr.api.service
```
suzannehamilton, line 1: Could you add or update a RouteSpec test, just to make sure things work end-to-end?

MancunianSam: I've added one to check for the metadata in `ConsignmentRouteSpec` I didn't want to add it to any of the existing ones as they're already pretty  busy.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
```diff
@@ -19,6 +20,7 @@ class DeferredResolver extends sangria.execution.deferred.DeferredResolver[Consi
       case DeferParentFolder(consignmentId) => context.consignmentService.getConsignmentParentFolder(consignmentId)
       case DeferConsignmentSeries(consignmentId) => context.consignmentService.getSeriesOfConsignment(consignmentId)
       case DeferConsignmentBody(consignmentId) => context.consignmentService.getTransferringBodyOfConsignment(consignmentId)
+      case DeferFile(consignmentId) => context.fileMetadataService.getFileMetadata(consignmentId)
```
suzannehamilton, line 23: Should this be plural? `DeferFiles`? Otherwise the name makes it sound like you fetch one file at once, but it actually returns a list.

suzannehamilton (commented):
Looks good! Just one comment about the deferred resolver name.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/DeferredResolver.scala
MancunianSam: Yep, that makes sense. I've changed it.

suzannehamilton (approved):

PR #120, opened 2021-02-09
Pass in export datetime instead of generating value

TomJKing:
Export datetime value is required before the DB table is updated, so client provides this value.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
@@ -13,7 +13,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class ConsignmentRepository(db: Database) {
 
   implicit class ZonedDateTimeUtils(value: ZonedDateTime) {
-    def toTimestamp = {
+    def toTimestamp: Timestamp = {
       Timestamp.valueOf(value.toLocalDateTime)
     }
   }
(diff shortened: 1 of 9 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -11,17 +12,23 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class ConsignmentRepository(db: Database) {
 
+  implicit class ZonedDateTimeUtils(value: ZonedDateTime) {
+    def toTimestamp: Timestamp = {
+      Timestamp.valueOf(value.toLocalDateTime)
+    }
```
MancunianSam, line 18: You've already got a `ZonedDateTimeUtils` implicit class in `FieldTypes` Is it worth merging this method into that?

TomJKing: Fair point, was thinking about this. I think it actually might be worth creating a new class `TimeUtils` and adding the implicit classes to this. Wasn't too comfortable putting this in the FieldTypes as these implicits can be used across the whole project, not just the graphql types. What do you reckon?

MancunianSam: Yeah that makes sense. Once my changes for the metadata go in there'll be LocalDateTime stuff so I can put them in `TimeUtils` as well.

MancunianSam (approved):
:+1:

MancunianSam (approved):

PR #121, opened 2021-02-09
Add missing type annotations in fields, rename test file and made …

techncl:


suzannehamilton (approved):
Thanks for fixing these!

PR #122, opened 2021-02-10
Fix consignment seq reference

Lholbourn:
Due to the production and test DB's being different types of database, the consignment sequence object needed it's name changed to follow PostgreSQL convention.

The previous casing of the sequence ("ConsignmentSequenceId") meant that the sql query needed specific quotation marks to override the fact that tables/objects/columns are usually in lower snake case within PostgreSQL DB's, but these meant that the H2 database detected the sql different, meaning either the production API worked OR the tests passed. 

This change will hopefully lead to both working in production and locally :+1:

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
@@ -42,7 +42,7 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
   }
 
   def getNextConsignmentSequence(implicit executionContext: ExecutionContext): Future[Long] = {
-    val query = sql"select nextval('ConsignmentSequenceID')".as[Long]
+    val query = sql"select nextval('consignment_sequence_id')".as[Long]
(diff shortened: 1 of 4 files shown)


MancunianSam (dismissed):
LGTM

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -42,7 +42,7 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
   }
```
TomJKing, line 42: There is an unused import (line 7): `import slick.sql.SqlStreamingAction`

TomJKing (commented):
Looks good. Just one very minor comment.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
Lholbourn: Ah I'll remove that :+1:

TomJKing (approved):

PR #123, opened 2021-02-16
Get consignment reference

Lholbourn:
These changes retrieve a specified consignment's consignment reference from the DB and creates a service to call this method.
A `consignmentReference` field has been added to the consignment fields and the deferred resolver. This will allow us to add the `consignmentReference` to the generated [graphql query](https://github.com/nationalarchives/tdr-generated-graphql/blob/master/src/main/graphql/GetConsignmentSummary.graphql) for the transfer summary page.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
@@ -88,6 +88,11 @@ object ConsignmentFields {
         "files",
         ListType(FileType),
         resolve = context => DeferFiles(context.value.consignmentid)
+      ),
+      Field(
+        "consignmentReference",
+        OptionType(StringType),
+        resolve = context => DeferConsignmentReference(context.value.consignmentid)
       )
     )
   )
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
@@ -95,4 +95,9 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
     val query = Consignment.filter(_.consignmentid === consignmentId).map(_.parentfolder)
     db.run(query.result).map(_.headOption.flatten)
   }
+
+  def getConsignmentReference(consignmentId: UUID)(implicit executionContext: ExecutionContext): Future[Option[String]] = {
+    val query = Consignment.filter(_.consignmentid === consignmentId).map(_.consignmentreference)
+    db.run(query.result).map(_.headOption.flatten)
+  }
 }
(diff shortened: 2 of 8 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -88,6 +88,11 @@ object ConsignmentFields {
         "files",
         ListType(FileType),
         resolve = context => DeferFiles(context.value.consignmentid)
+      ),
+      Field(
+        "consignmentReference",
+        OptionType(StringType),
+        resolve = context => DeferConsignmentReference(context.value.consignmentid)
```
suzannehamilton, line 95: Is there a particular reason for making this a deferred value, rather than a regular field like `userid`?

Do you think we should have some guidance in the Readme about when to use deferred resolvers?

Lholbourn: I thought since it's only going to be used for one page, we might not need to have the consignment reference every time we call `getConsignment`. How much would change if I did it as a regular field instead?

I think some sort of guidance would be good yeah, I'll be honest I'm not 100% sure when I should use them and when not to 😅

suzannehamilton: The consignment reference is a column in the consignment table, so if it was a regular field then it would be fetched with all the rest of the regular fields in `getConsignment`. This will just add one column to the database query. It will go from something like:

```
SELECT ConsignmentId, SeriesId, Datetime, ParentFolder FROM Consignment
```

to:


```
SELECT ConsignmentId, SeriesId, Datetime, ParentFolder, ConsignmentRefrence FROM Consignment
```

I wouldn't expect that to cause a performance problem for other clients calling `getConsignment` but which don't need the series reference.

On the other hand, if it's a deferred resolver, a query to `getConsignment that includes the consignment reference will have to make two database calls rather than one. To be honest, I wouldn't expect that to be very slow either, but if we keep repeating that pattern then it will get out of hand because we'll end up making a database query for each column separately. 🙂 

I'll make a note about adding some general guidance on deferred resolvers to the Readme. 👍

Lholbourn: Ah awesome, thank you for the info! 👼 
Do you think it's more sensible to change it to a normal field in this case then?

suzannehamilton: I think so, if it's not too big a change?

I'm going to approve this PR so we can merge it as it currently is, since the code looks sensible and there are a couple of people who are waiting for this field. Then would you mind raising a separate PR to make it a regular field, assuming that it's not too difficult to change? Hopefully the route specs won't change.

suzannehamilton (approved):

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
Lholbourn: Sure, I'll get on it ASAP! 👍

Lholbourn: Will changing this to a normal field mean that the [generated GraphQL query](https://github.com/nationalarchives/tdr-generated-graphql/pull/88) will have to be modified as well or should that not be affected by this change?

suzannehamilton: It shouldn't change the query. If I remember correctly, whether a field is deferred or not is an implementation detail that's invisible to the API clients.

PR #124, opened 2021-02-23
Make consignment ref normal field

Lholbourn:
This PR changes the `consignmentReference` field on the API to be a normal field rather than utilising the deferred resolver. It has no effect on the [relevant generated graphql query](https://github.com/nationalarchives/tdr-generated-graphql/blob/master/src/main/graphql/GetConsignmentSummary.graphql#L10).

I also took the chance to fix some of the routeSpec tests as they were explicitly inserting into the test `Consignment` table rather than utilising available helper methods. I edited the existing helper method and changed tests to call this rather than hard-coding SQL. Though there was a test within the `ConsignmentRouteSpec` which was inserting more complex data that I left hard-coded.


suzannehamilton (approved):
Looks good, and thanks for commonising all that test setup! 🚀

PR #125, opened 2021-03-11
Iam db auth

MancunianSam:
Each time we get a query, we create the db object from the SlickSession as before. But now we create a token using the rds sdk and set the database password to this. This is then stored in an in memory cache. Each subsequent query for 13 minutes will use this cached password. Once the cache expires, we make the call to the rds sdk again and update the password with the new token.
I've tested this with the intg database through an ssh tunnel and you can connect using IAM auth with two queries more than 15 minutes apart.


src/main/resources/application.base.conf
```diff
@@ -13,12 +13,13 @@ source = {
...
 consignmentapi = {
+  useIamAuth = true
   profile = "slick.jdbc.PostgresProfile$"
   db {
     connectionPool = "HikariCP"
     driver = "org.postgresql.Driver",
     url = "jdbc:postgresql://"${DB_ADDR}":"${DB_PORT}"/consignmentapi?ssl=true&sslrootcert=/api/rds-ca-2019-root.pem&sslmode=verify-full",
     user = ${DB_USER}
-    password = ${DB_PASSWORD}
+    registerMbeans = true
```
suzannehamilton, line 23: What does this line do?

suzannehamilton (commented):
This looks like it's worth doing, thanks! Is there anything else that you'd want to check before we go down this route?

Once we've decided that's what we'll do, I'll let you do any tidying up you wanted to do before I add any specific review comments.

src/main/resources/application.base.conf
MancunianSam: We're setting the password using `hikariDataSource.ds.getHikariConfigMXBean.setPassword` This config option is needed to allow you to do it. https://github.com/brettwooldridge/HikariCP/wiki/MBean-(JMX)-Monitoring-and-Management

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
```diff
@@ -1,7 +1,57 @@
...
+
+  def getPassword: Try[String] = memoize[Try, String](Some(5.minutes)) {
+    val configFactory = ConfigFactory.load
+    val useIamAuth = configFactory.getBoolean("consignmentapi.useIamAuth")
+    if (useIamAuth) {
+      val rdsClient = RdsUtilities.builder().region(Region.EU_WEST_2).build()
+      val port = 5432
+      val request = GenerateAuthenticationTokenRequest.builder()
+        .credentialsProvider(DefaultCredentialsProvider.builder().build())
+        .hostname(sys.env("DB_ADDR"))
```
suzannehamilton, line 47: Could this be read from an environment variable in the config, like the rest of the env vars?

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
```diff
@@ -1,7 +1,57 @@
...
+        db
+    }
+  }
+
+  def getPassword: Try[String] = memoize[Try, String](Some(5.minutes)) {
+    val configFactory = ConfigFactory.load
+    val useIamAuth = configFactory.getBoolean("consignmentapi.useIamAuth")
+    if (useIamAuth) {
+      val rdsClient = RdsUtilities.builder().region(Region.EU_WEST_2).build()
+      val port = 5432
```
suzannehamilton, line 44: This looks like it should be in the config. I know it's the default Postgres port, but does it technically come from Terraform?

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
```diff
@@ -1,7 +1,57 @@
...
+      val request = GenerateAuthenticationTokenRequest.builder()
+        .credentialsProvider(DefaultCredentialsProvider.builder().build())
+        .hostname(sys.env("DB_ADDR"))
+        .port(port)
+        .username("api_iam")
```
suzannehamilton, line 49: This looks like it should also be in the config file for consistency. Could it be clear that it's the consignment API? "API" could mean anything if you see it out of context.

src/main/resources/application.base.conf
suzannehamilton: Probably worth a comment, since it's not very obvious.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
MancunianSam: Yeah makes sense. The only reason I didn't originally was because there was no parameter in application.conf with just this value, it was used as part of the url. I've changed application.conf to have a host property which I'm reading in here.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
MancunianSam: I've changed this one. We already get it in as an environment variable so I've added a property to application.conf

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
MancunianSam: I've changed it to `consignment_api_user` and I've added it to the config file as `newUser` If I change the existing username in the environment variables then we'll get down time between running terraform and deploying the API so I've added a new property. Once it's deployed and working, we can get rid of `newUser` and update `user` to be the new user.

src/main/resources/application.base.conf
MancunianSam: I've added one in.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/DbConnection.scala
```diff
@@ -1,7 +1,57 @@
...
+          case Success(password) =>
+            configBean.setPassword(password)
+            db
+        }
+      case _ =>
+        db
+    }
+  }
+
+  def getPassword: Try[String] = memoize[Try, String](Some(5.minutes)) {
```
suzannehamilton, line 39: Sorry, just one more suggestion for a comment: could mention that this is chosen to be less than the 15 minute expiry and link to a doc about temporary IAM passwords?

MancunianSam: I've added a comment in.

suzannehamilton (approved):

PR #126, opened 2021-03-12
Add ffid metadata fields to existing API endpoint

techncl:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -48,6 +49,49 @@ class FileService(
   def getFiles(consignmentId: UUID): Future[Files] = {
     fileRepository.getFilesWithPassedAntivirus(consignmentId).map(r => Files(r.map(_.fileid)))
   }
+
+  def getFileMetadata(consignmentId: UUID): Future[List[File]] = {
```
suzannehamilton, line 53: This is an interesting one. The method is now fetching a lot of different data, so we should probably use deferred resolvers, but it's already inside another deferred resolver and I've never tried nesting deferred resolvers in Sangria.

It looks like you _can_ do it, though: https://github.com/sangria-graphql/sangria/issues/63

I don't think it's going to be trivial to rewrite this code with nested resolvers, but we should do it because this is going to get even more complicated when we add the antivirus metadata. So I think I'll add a refactoring card for us to dig into this properly, and stick with what you've got here. How does that sound?

suzannehamilton (commented):
Thanks for being so thorough about what you're testing. 👍 

A couple of suggestions, and a note about some future refactoring of how we fetch file data.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
suzannehamilton: I've made a card about it: https://national-archives.atlassian.net/browse/TDR-948

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
```diff
@@ -83,13 +83,93 @@ class FFIDMetadataRepositorySpec extends AnyFlatSpec with TestDatabase with Scal
     TestUtils.createFile(UUID.fromString(fileThreeId), consignmentId)
     TestUtils.createFile(UUID.fromString(fileFourId), consignmentId)
 
-    (1 to 7).foreach { _ => TestUtils.addFFIDMetadata(fileOneId)}
+    TestUtils.addFFIDMetadata(fileOneId)
+    TestUtils.addFFIDMetadata(fileTwoId)
+    TestUtils.addFFIDMetadata(fileThreeId)
```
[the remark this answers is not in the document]
techncl: Ah, yes, because I noticed the same test in the FileMetadata Repository and thought that it was a just copied over by mistake, because I was wondering why you would you call addFFIDMetadata multiple times on a file, but I guess it could happen and probably might since the fileId doesn't have to be unique which also explains why we have a `groupBy` in the "count" slick query. 

1. Are we allowing multiple FFID metadata for versioning purposes (i.e. "this file was updated with a new FFID on this date, originally it was X")? 
      i. Could this same logic apply to the File metadata table? Could you/would you want to have multiple client-side checksum rows? Maybe if the process of updating the database fails and you need to start again? 
2. Or can a file have multiple file format identifications for a single file?

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
```diff
@@ -83,13 +83,93 @@ class FFIDMetadataRepositorySpec extends AnyFlatSpec with TestDatabase with Scal
...
+
+    TestUtils.addFFIDMetadata(fileOneId)
     TestUtils.addFFIDMetadata(fileTwoId)
     TestUtils.addFFIDMetadata(fileThreeId)
 
-    val fileMetadataFiles = ffidMetadataRepository.countProcessedFfidMetadata(consignmentId).futureValue
+    val ffidMetadataRows = ffidMetadataRepository.getFFIDMetadata(consignmentId).futureValue
 
-    fileMetadataFiles shouldBe 3
+    ffidMetadataRows.toMap.values.foreach(
```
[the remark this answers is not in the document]
techncl: I just realised that this test in particular is testing whether the IDs of the FFIDMetadata and the FFIDMetadataMatches match each other (i.e. the Slick logic is sound), should I then: 

1. Do a test of equality on the two Ids and then test if one of those Ids is the same as the one I sent in

    ffidMetadataRows.toMap.values.foreach(
      ffidMetadataAndMatches =>
        ffidMetadataAndMatches._1.ffidmetadataid == ffidMetadataAndMatches._2.ffidmetadataid should equal(true)
        ffidMetadataAndMatches._1.ffidmetadataid == <ffimetadataId>
        
        OR simply 

        ffidMetadataAndMatches._1.ffidmetadataid == <ffimetadataId>
        ffidMetadataAndMatches._2.ffidmetadataid == <ffimetadataId>
    )

    To make what I'm testing clearer I could also save the tuple values to variables, e.g.
    
        ffidMetadataRows.toMap.values.foreach{
          ffidMetadataAndMatches =>
            val ffidMetadata = ffidMetadataAndMatches._1.ffidmetadataid
            val ffidMetadataMatches = ffidMetadataAndMatches._2.ffidmetadataid
            ffidMetadata.ffidmetadataid == ffidMetadataMatches.ffidmetadataid should equal(true)
            ffidMetadata.ffidmetadataid == <ffimetadataId>
        }

2. And then create another test that tests if what comes out is what I put in, like I did for the [FileService tests](https://github.com/nationalarchives/tdr-consignment-api/blob/8784e2156789632d2297f82e4ce270a7671d90fb/src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala#L216)?

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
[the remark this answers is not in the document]
suzannehamilton: It's a good question, and one we don't really have an answer to yet.

I can only think of two ways that you might end up with multiple FFID rows for the same file:

1. There's some network glitch when the api-update lambda saves the metadata to the consignment API which leads to the data being saved but the api-update lambda not receiving the successful response. The api-update lambda tries up to three times to save the data, so it will assume the first one has failed and try again. In this case, the data in each row will be identical.
2. We re-trigger the file check messages for a file, even though the initial file checks succeeded. We don't have a mechanism for doing this at the moment, but I suspect we'll need to build one at some point. In this case, the data could potentially be different if, for example, we upgraded DROID in between running the first checks and the second.

So there is a bit of ambiguity here which we should definitely sort out at some point, but I don't think we have enough information to know what we _should_ do here. It will depend on what information we want to keep for audit, and on what we think we'll do when we let users replace specific files (e.g. letting them overwrite a password-protected file with the unlocked file). So I'd rather tackle this if we end up in situation 2 a lot, or when we start tackling file replacement.

In the meantime, I think it's worth keeping the old test logic which saves some duplicate FFID so that we understand what happens when there's duplicate data, even though we don't think it will happen very often.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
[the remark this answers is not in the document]
suzannehamilton: Let's have a chat about this on a call so we can clarify what we actually want to test.

suzannehamilton:
Just noticed that there are a couple of tests failing on Jenkins.

techncl:
> Just noticed that there are a couple of tests failing on Jenkins.

Yes, I'm looking into what went wrong. I merged another branch (v1 of this one) that I accidentally committed to, into this one and I think it seems as though I broke something in another test.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
```diff
@@ -58,16 +58,173 @@ class FFIDMetadataRepositorySpec extends AnyFlatSpec with TestDatabase with Scal
...
+    val fileThreeFfidMetadataId = TestUtils.addFFIDMetadata(fileThreeId)
+    TestUtils.addFFIDMetadataMatches(fileThreeFfidMetadataId.toString)
+
+    val fileFourFfidMetadataId = TestUtils.addFFIDMetadata(fileFourId)
+    TestUtils.addFFIDMetadataMatches(fileFourFfidMetadataId.toString)
+
+    val ffidMetadataRows = ffidMetadataRepository.getFFIDMetadata(consignmentOneId).futureValue
+    val fileIds: Set[UUID] = ffidMetadataRows.toMap.keySet
+
+    fileIds.contains(UUID.fromString(fileOneId)) should equal(true)
```
suzannehamilton, line 182: Minor thing, but these assertions could be written as:

```
fileIds should contain(UUID.fromString(fileOneId))
```

(Or `should not contain(foo)` in the final case.)

It means that if it fails, you get an error message like:

```
Set(50f4f290-cdcc-4a0f-bd26-3bb40f320b71, 7f55565e-bfa5-4cf9-9e02-7e75ff033b3b, 89b6183d-e761-4af9-9e37-2ffa09922dba) did not contain element 939999e5-c082-4052-a99a-6b45092d826f
```

Rather than:

```
false did not equal true
```

The former is easier to debug because you have more information about what the assertion was looking for.

It's a good principle to follow in general: if a test assertion is "something should equal true" (or false), then you can usually find a better assertion in the testing library's list of assertions. The main exception is if the function you're testing returns a simple boolean, because then the only thing you _can_ test is whether the result is true or false.

suzannehamilton: Same goes for the next test as well.

suzannehamilton (commented):
Just one small suggestion for improving a test assertion. 👍

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FFIDMetadataRepositorySpec.scala
techncl: Ah, cool, I didn't know about this; thanks for tip. It also reads better/more naturally 👍

suzannehamilton (approved):

PR #128, opened 2021-03-24
Add IntelliJ test run instructions to ensure devs can effectively run tests

Lholbourn:
IntelliJ is confusing run config with test run config and causing issues when trying to run the API tests. These instructions should help other developers run tests for API code without coming across this error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,17 @@ Run the api:
 * From IntelliJ, run the ApiServer app
 * Or from the command line, run `sbt run`
 
+### Testing within IntelliJ
+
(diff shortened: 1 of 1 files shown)


suzannehamilton:
What's happening with this PR? Did the discussion on Slack come up with a different solution?

Lholbourn:
> What's happening with this PR? Did the discussion on Slack come up with a different solution?

No other solution seemed to appear from the discussion on Slack, so I think this might be the best solution for now 👍

suzannehamilton (approved):

suzannehamilton (approved):

PR #130, opened 2021-04-06
FFID match changes

MancunianSam:
There's two things here. The first is to make ffidMetadata optional as if you try to query it when it isn't set, you get an internal server error.

The second is to allow it to return multiple matches. The original code was calling `ffidMetadataAndMatchesRows.toMap` but if you had more than one match, it was trying to create to entries in that map with the same key which you can't have so you ended up with a random ffid match.


src/test/resources/application.conf
```diff
@@ -21,6 +21,7 @@ consignmentapi = {
     driver = org.h2.Driver
     connectionPool = disabled
     keepAliveConnection = true
+    port = 5432
```
suzannehamilton, line 24: @Lholbourn  added this as an alternative solution last week: https://github.com/nationalarchives/tdr-consignment-api#testing-within-intellij

Which do you think we should go with?

suzannehamilton (commented):
Main change looks good, thanks. Just one query about the DB port config, because it looks like there are two different solutions and we should probably pick one.

src/test/resources/application.conf
MancunianSam: Lets go with setting the -Dconfig.file variable. It doesn't involve introducing variables that don't do anything like this one does. I've taken these lines out.

suzannehamilton (approved):

PR #132, opened 2021-04-19
Refactor file service getFileMetadata method

MancunianSam:
The getFileMetadata method was getting complicated because all the logic for getting the FFID and client side metadata was in a single method. If we were to add the antivirus to this then it would be too messy. I've refactored this so the client metadata and the FFID metadata comes from different methods.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
@@ -1,8 +1,7 @@
 package uk.gov.nationalarchives.tdr.api.service
 
-import uk.gov.nationalarchives.Tables.{ConsignmentstatusRow, FfidmetadataRow, FfidmetadatamatchesRow, FileRow, FilemetadataRow}
+import uk.gov.nationalarchives.Tables.{ConsignmentstatusRow, FileRow, FilemetadataRow}
 import uk.gov.nationalarchives.tdr.api.db.repository._
-import uk.gov.nationalarchives.tdr.api.graphql.fields.FFIDMetadataFields.{FFIDMetadata, FFIDMetadataMatches}
 import uk.gov.nationalarchives.tdr.api.graphql.fields.FileFields.{AddFilesInput, Files}
 import uk.gov.nationalarchives.tdr.api.service.FileMetadataService._
 
@@ -15,6 +14,7 @@ class FileService(
                    consignmentRepository: ConsignmentRepository,
                    fileMetadataRepository: FileMetadataRepository,
                    ffidMetadataRepository: FFIDMetadataRepository,
+                   ffidMetadataMatchesRepository: FFIDMetadataMatchesRepository,
                    timeSource: TimeSource,
                    uuidSource: UUIDSource
                  )(implicit val executionContext: ExecutionContext) {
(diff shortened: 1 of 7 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -53,45 +53,17 @@ class FileService(
...
-                  matches.map(m => FFIDMetadataMatches(m.extension, m.identificationbasis, m.puid)).toList,
-                  metadata.datetime.getTime
-                )
-              }
-              File(fileId, fileMetadataValues, fullFfidMetadata)
-            }).toList
-        }
-    }.flatten
+    val fileMetadataService = new FileMetadataService(fileMetadataRepository, timeSource, uuidSource)
+    val ffidMetadataService = new FFIDMetadataService(ffidMetadataRepository, ffidMetadataMatchesRepository, timeSource, uuidSource)
```
suzannehamilton, line 57: Could these be passed to this service's constructor, so that all the services are created in the same place?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataServiceSpec.scala
```diff
@@ -81,6 +83,82 @@ class FFIDMetadataServiceSpec extends AnyFlatSpec with MockitoSugar with Matcher
...
+
+    when(ffidMetadataRepositoryMock.getFFIDMetadata(consignmentIdCaptor.capture())).thenReturn(Future(Seq()))
+
+    val service = new FFIDMetadataService(ffidMetadataRepositoryMock, mock[FFIDMetadataMatchesRepository], FixedTimeSource, new FixedUUIDSource())
+    service.getFFIDMetadata(consignmentId)
+
+    consignmentIdCaptor.getValue should equal(consignmentId)
+  }
+
+  "getFFIDMetadata" should "return the correct values" in {
```
suzannehamilton, line 99: Could this test name be more specific than "correct values"? What are you expecting - FFID metadata for all files belonging to the given consignment?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
```diff
@@ -140,4 +140,39 @@ class FileMetadataServiceSpec extends AnyFlatSpec with MockitoSugar with Matcher
...
+    val mockResponse = Future(Seq())
+
+    when(fileMetadataRepositoryMock.getFileMetadata(consignmentIdCaptor.capture())).thenReturn(mockResponse)
+
+    val service = new FileMetadataService(fileMetadataRepositoryMock, FixedTimeSource, new FixedUUIDSource())
+    service.getFileMetadata(consignmentId).futureValue
+    consignmentIdCaptor.getValue should equal(consignmentId)
+  }
+
+  "getFileMetadata" should "return the correct values for multiple files" in {
```
suzannehamilton, line 157: Same comment - it makes the test easier to read if you can be more specific than "correct".

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
MancunianSam: The only reason I didn't is because there's an existing method in `FileService` `addFile` which is directly using the `FileMetadataRepository` and I didn't want the service and the repository in the same constructor. What makes more sense now you say it is to add another method to `FileMetadataService` to add the static metadata and then all we need to have in the `FileService` constructor is the `FileMetadataService` and `FFIDMetadataService`
If this was a Java project I'd say this is facade territory but I think this is ok.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataServiceSpec.scala
MancunianSam: I've changed it to `return multiple rows and matches for multiple files` which is a bit more descriptive but I'm open to suggestions.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
MancunianSam: I've put `return mutiple map entries for multiple files` which I guess is more descriptive.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataServiceSpec.scala
MancunianSam: Sorry, I hadn't pushed the change.

suzannehamilton (approved):

PR #135, opened 2021-04-27
Create checksum file status

MancunianSam:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
```diff
@@ -0,0 +1,26 @@
...
+    for {
+      checksumStatus <- fileStatusRepository.getFileStatus(consignmentId, "checksum")
+    } yield checksumStatus.headOption.exists(_.value == Success)
+  }
+
+}
+
+object FileStatusService {
+  //Status types
+  val Checksum = "checksum"
```
suzannehamilton, line 21: This should be capitalized: `Checksum`. That's my fault, sorry - I copied it manually to the Jira ticket.

src/test/resources/json/addfilemetadata_data_all.json
```diff
@@ -3,7 +3,7 @@
     "addFileMetadata": {
       "filePropertyName": "SHA256ServerSideChecksum",
       "fileId": "07a3a4bd-0281-4a6d-a4c1-8fa3239e1313",
-      "value": "value"
+      "value": "SHA256ClientSideChecksum Value"
```
suzannehamilton, line 6: This says `ClientSide` but line 4 is `ServerSide`. Copy-paste error?

suzannehamilton: Ah, or is it this so that it matches the client-side value? I wonder if it would be better to rename it to make it easier to follow.

MancunianSam: It needs to be `SHA256ServerSideChecksum` as the property name because it's mocking the response getting that value. It is a bit weird having the value set to that though. I've changed it to a random checksum in all the places it's needed. What do you think?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
MancunianSam: It's ok. I've changed it.

suzannehamilton (approved):
Looks good, thanks!

PR #138, opened 2021-05-11
Tdr 1062 update file status with file check

TomJKing:


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FFIDMetadataRouteSpec.scala
```diff
@@ -58,6 +59,16 @@ class FFIDMetadataRouteSpec extends AnyFlatSpec with Matchers with TestRequest w
...
+  "addFFIDMetadata" should "set the file status to password protected found when a password protected file is found" in {
+    runTestMutation("mutation_password_protected", validBackendChecksToken("file_format"))
+    getFileStatusResult(defaultFileId, FFID) should equal(PasswordProtected)
+  }
+
+  "addFFIDMetadata" should "set the file status to success when there are no password protected files found" in {
+    runTestMutation("mutation_alldata", validBackendChecksToken("file_format"))
+    getFileStatusResult(defaultFileId, FFID) should equal(Success)
+  }
+
```
MancunianSam, line 71: Is it worth adding a test where we've got one match which shows password protected and another that shows a valid puid? I got tripped up with multiple matches on the last one and it might break later as we add more invalid puids

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
```diff
@@ -51,6 +63,13 @@ class FFIDMetadataService(ffidMetadataRepository: FFIDMetadataRepository, matche
     }
   }
 
+  private def checkForPasswordProtectedPuids(matchesMetadata: FFIDMetadataInput): String = {
+    matchesMetadata.matches.map(_.puid.getOrElse("")).exists(passwordProtectedPuids.contains) match {
+      case true => PasswordProtected
+      case _ => Success
+    }
+  }
+
```
MancunianSam, line 72: IntelliJ is complaining here because it doesn't like `match` statements with only true and false and prefers if/else. This might be a bit of premature optimisation but would it be worth making it a bit more generic, something like
```scala
private def checkForInvalidPuids(matchesMetadata: FFIDMetadataInput): String = {
    matchesMetadata.matches.map(_.puid.getOrElse("")) match {
      case puids if puids.exists(passwordProtectedPuids.contains) => PasswordProtected
      //we can add more cases here when we're filtering out other files
      case _ => Success
    }
  }
```
Or that might just be confusing and it would make more sense to have them as separate methods. I don't mind so feel free to ignore me.

TomJKing: Make sense to me

MancunianSam (approved):

PR #140, opened 2021-05-14
Set json logs and output file status

MancunianSam:
- Set logs to output in json format.
- Log the file status so we can query this in Grafana

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
@@ -1,20 +1,24 @@
 package uk.gov.nationalarchives.tdr.api.service
 
+import com.typesafe.scalalogging.Logger
 import uk.gov.nationalarchives
 import uk.gov.nationalarchives.Tables
+
 import java.sql.{SQLException, Timestamp}
 import java.util.UUID
-
 import uk.gov.nationalarchives.Tables.{FfidmetadataRow, FfidmetadatamatchesRow, FilestatusRow}
 import uk.gov.nationalarchives.tdr.api.db.repository.{FFIDMetadataMatchesRepository, FFIDMetadataRepository}
 import uk.gov.nationalarchives.tdr.api.graphql.DataExceptions.InputDataException
 import uk.gov.nationalarchives.tdr.api.graphql.fields.FFIDMetadataFields.{FFIDMetadata, FFIDMetadataInput, FFIDMetadataMatches}
 import uk.gov.nationalarchives.tdr.api.service.FileStatusService.{FFID, PasswordProtected, Success, Zip}
+import net.logstash.logback.argument.StructuredArguments._
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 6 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
```diff
@@ -35,6 +39,11 @@ class FFIDMetadataService(ffidMetadataRepository: FFIDMetadataRepository, matche
 
     val fileStatusRows = generateFileStatusRows(ffidMetadata)
 
+    logger.info("File check {} for fileId {} completed with status {}",
```
suzannehamilton, line 42: How does this appear in the logs? It looks like there's some string interpolation going on, but do the values below end up in separate JSON fields as well for querying?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -46,6 +51,12 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
             case None => throw new IllegalStateException(s"Cannot find client side checksum for file ${addFileMetadataInput.fileId}")
           }
           fileStatusRow: FilestatusRow = FilestatusRow(uuidSource.uuid, addFileMetadataInput.fileId, Checksum, fileStatus, timestamp)
+          _ <- Future {
+            logger.info("File check {} for fileId {} completed with status {}",
```
suzannehamilton, line 55: If the structure of these three messages needs to be kept consistent for querying, then could you extract a shared method to do it?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
MancunianSam: Yep, you get the string as a message but you also get the fields included separately. This is an example that I got from running the tests.
```json
{
   "@timestamp":"2021-05-17T11:04:55.433+01:00",
   "@version":"1",
   "message":"File check FFID for fileId 07a3a4bd-0281-4a6d-a4c1-8fa3239e1313 completed with status Success",
   "logger_name":"FFIDMetadataService",
   "thread_name":"consignmentApi-akka.actor.default-dispatcher-6",
   "level":"INFO",
   "level_value":20000,
   "fileCheck":"FFID",
   "fileId":"07a3a4bd-0281-4a6d-a4c1-8fa3239e1313",
   "fileCheckStatus":"Success",
   "application":"consignment-api"
}
```

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: Yeah, makes sense. I've created a LoggingUtils class and moved it in there.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
suzannehamilton: That's really nice!

suzannehamilton (approved):

PR #141, opened 2021-05-17
TDR-240 Return the same error for non existent consignment.

MancunianSam:
When we validate that the user owns the consignment, if the consignment
doesn't exist, we return "Invalid consignment Id" and if they don't have
permissions, then we get user xxx does not have accesss...
It's good practice to return the same authorisation error message
whether a resource exists or not to prevent people finding out which
resources exist by checking for the errors. I've changed it so you get
the same message and updated the tests which were checking for it.


src/test/resources/json/addfinaltransferconfirmation_data_error_invalid_consignmentid.json
```diff
@@ -2,7 +2,7 @@
   "data": null,
   "errors": [
     {
-      "message": "Invalid consignment id"
+      "message": "User '4ab14990-ed63-4615-8336-56fbb9960300' does not have access to consignment 'f67d1337-cbd0-4fd1-9eac-611489e7113f'"
     }
   ]
 }
```
TomJKing, line 8: Missing line break

src/test/resources/json/addtransferagreement_data_error_invalid_consignmentid.json
```diff
@@ -2,7 +2,7 @@
   "data": null,
   "errors": [
     {
-      "message": "Invalid consignment id"
+      "message": "User '4ab14990-ed63-4615-8336-56fbb9960300' does not have access to consignment 'f67d1337-cbd0-4fd1-9eac-611489e7113f'"
     }
   ]
 }
```
TomJKing, line 8: Missing line break

TomJKing (commented):
Looks good. Couple of very minor comments.

src/test/resources/json/addfinaltransferconfirmation_data_error_invalid_consignmentid.json
MancunianSam: I've been trying with the line breaks :smile: I've added this one and the one below.

TomJKing (approved):

PR #143, opened 2021-05-19
Rename body code to tdr code

MancunianSam:
- Rename code to tdrCode for Body
- Update consignment data version


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -63,7 +63,7 @@ class ConsignmentService(
   def getTransferringBodyOfConsignment(consignmentId: UUID): Future[Option[TransferringBody]] = {
     val consignment: Future[Seq[BodyRow]] = consignmentRepository.getTransferringBodyOfConsignment(consignmentId)
     consignment.map(rows => rows.headOption.map(
-      transferringBody => TransferringBody(transferringBody.name, transferringBody.code)))
+      transferringBody => TransferringBody(transferringBody.name, transferringBody.tdrcode)))
```
suzannehamilton, line 66: Sometimes this is `tdrcode` and sometimes it's `tdrCode`. Is it possible to be consistent, or is this something about the way the code is auto-generated from the DB?

MancunianSam: Yeah, it's the auto-generated code does the field names all in lower case. I suspect it's because we're using pascal case for the column names rather than snake case which is the standard for postgres but I'm not sure. So it's `tdrcode` where I've got no control over it and `tdrCode` for case classes and things where I can set them.

suzannehamilton (approved):

PR #144, opened 2021-05-25
Add graphql schema file and check

MancunianSam:
I've added in a `schema.graphql` file which shows the current API schema and an sbt task called compareSchema.

This will run graphqlSchemaGen and compare the output with the file in `schema.graphql`. If they don't match, then the test build will fail. I've updated the README to include instructions on what to do.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,7 @@ import sbt.File
 import sbt.Keys.libraryDependencies
 
 import java.nio.file.Files
+import scala.io.Source
 
 
 name := "tdr-consignment-api"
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -19,6 +19,16 @@ resolvers ++= Seq[Resolver](
 
 mainClass in (Compile, run) := Some("uk.gov.nationalarchives.tdr.api.http.ApiServer")
 
+val compareSchema = taskKey[Unit]("Generate the graphql schema and compares it with schema.graphql in the project root")
+
+compareSchema := {
```
suzannehamilton, line 24: Just wondering if you saw the `graphqlValidateSchema` task in sbt-graphql, and whether it's worth using it since it looks like it might compare the schemas logically rather than byte-for-byte? https://github.com/muuki88/sbt-graphql#schema-comparison

MancunianSam: Ah good spot yeah. I've added the schema from the file to build sbt and then you can compare them using the `graphqlValidateSchema` It's better because it will tell you which fields have been added or removed rather than just telling you they're different.

suzannehamilton (approved):
Looks good! 👍

PR #146, opened 2021-06-03
Update generated graphql version

MancunianSam:
This includes the change to drop the Code column from the Body table which will stop slick trying to select from it.
The new changes also include making some columns mandatory. To update to the latest version of the generated slick classes, I've removed all of the options from these fields.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
@@ -56,8 +56,8 @@ object ValidateSeries extends AuthorisationTag {
 
     bodyResult.map(body => {
       body.tdrCode match {
-        case Some(code) if code == userBody => continue
-        case Some(code) =>
+        case code if code == userBody => continue
+        case code if code != userBody =>
           val message = s"User '${token.userId}' is from transferring body '$userBody' and does not have permission " +
             s"to create a consignment under series '$addConsignmentInput' owned by body '$code'"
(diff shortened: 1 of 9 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -56,8 +56,8 @@ object ValidateSeries extends AuthorisationTag {
 
     bodyResult.map(body => {
       body.tdrCode match {
-        case Some(code) if code == userBody => continue
-        case Some(code) =>
+        case code if code == userBody => continue
+        case code if code != userBody =>
```
suzannehamilton, line 60: Is this `if` statement necessary? Won't it work like before if it's not there?

suzannehamilton (commented):
Just one small comment.

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
MancunianSam: Because `String` can only be a `String` unlike `Option` which can be `Some` or `None`, line 60 will catch anything that isn't matched by line 59. This makes the `case _ => throw new IllegalStateException("")` redundant because you can never get there.
What I've realised now is that field is now mandatory so it never won't be populated, so we don't need the `case _ => throw new IllegalStateException("")` case and so we don't need the new if statement either. I've taken them both out.

suzannehamilton (approved):

PR #150, opened 2021-06-10
Add strict transport security header

MancunianSam:
I've added in the strict-transport-security header to the routes.


suzannehamilton (approved):
Sorry, looks like I just commented last time. I forgot to approve!

PR #151, opened 2021-06-10
Add 'file' endpoint to retrieve all file data

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -47,6 +48,13 @@ class FileService(
     fileRepository.getFilesWithPassedAntivirus(consignmentId).map(r => Files(r.map(_.fileid)))
   }
 
+  def getFile(fileId: UUID): Future[FileDetails] = {
+    fileRepository.getFileIfPassedAntivirus(fileId).map(r => {
```
suzannehamilton, line 52: Do we always want to apply this filter? It seems like it would be clearer to make it an explicit filter in the query.

Also, do we even need to do this filtering any more? Or did it come from a point when we thought we'd continue with the transfer but skip the file that failed antivirus? Now we're stopping the user from continuing with the consignment.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
```diff
@@ -28,6 +49,14 @@ object FileFields {
...
       resolve = ctx => ctx.ctx.fileService.getFiles(ctx.arg(ConsignmentIdArg)),
       tags=List(ValidateHasExportAccess)
+    ),
+    Field(
+      "file",
+      FileDetailsType,
+      arguments = List(FileIdArg),
+      resolve = ctx => ctx.ctx.fileService.getFile(ctx.arg(FileIdArg)),
+      //Assumption that only if the user owns the files can they view the file details
+      tags=List(ValidateUserOwnsFiles)
```
suzannehamilton, line 59: Will this endpoint be called by the frontend, or is it just a backend task? If it's the latter, could the authorisation be simpler because we just need to check for the file check roles in the token?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
```diff
@@ -1,25 +1,46 @@
...
 import uk.gov.nationalarchives.tdr.api.consignmentstatevalidation.ValidateNoPreviousUploadForConsignment
-import uk.gov.nationalarchives.tdr.api.graphql.ConsignmentApiContext
+import uk.gov.nationalarchives.tdr.api.graphql.{ConsignmentApiContext, DeferFileMetadata}
 import uk.gov.nationalarchives.tdr.api.graphql.validation.UserOwnsConsignment
 import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
+import uk.gov.nationalarchives.tdr.api.service.FileMetadataService.FileMetadataValues
 
 object FileFields {
   case class Files(fileIds: Seq[UUID])
+  case class FileDetails(fileId: UUID, consignmentId: UUID, userId: UUID, dateTime: LocalDateTime, checksumMatches: Option[Boolean])
```
suzannehamilton, line 19: Is there any reason you can't use the existing `File` type? https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala#L49

I think that would mean you'd only have one `File` type in the schema, whether you fetched `{ file { id } }` or `{ consignment { file { id } } }`.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
TomJKing: Was following what was done for the existing `getFiles` function which applies the anti-virus check.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
TomJKing: That `file` type does not contain all the information that could be retrieved about the file:

- consignmentId
- userId
- dateTime
- checksumMatches

Could potentially update the existing type to include to include this, but this will all have to be optional fields so as not to break existing code.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
TomJKing: At the moment the only places where the `getOriginalPath`, which has triggered TDR-874 task is in the:
- tdr-download-files
- tdr-local-aws

So essentially it is just the backend.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileFields.scala
suzannehamilton: Ah, so there are fields when we could return `None` just because we haven't implemented them? That doesn't seem great... Maybe we should have a chat about how we're approaching our GraphQL schema design?

PR #152, opened 2021-06-15
Tdr 1197 reporting consignments

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -124,6 +123,11 @@ object ConsignmentFields {
       arguments = ConsignmentIdArg :: Nil,
       resolve = ctx => ctx.ctx.consignmentService.getConsignment(ctx.arg(ConsignmentIdArg)),
       tags = List(ValidateUserHasAccessToConsignment(ConsignmentIdArg))
+    ),
+    Field("consignments", ListType(ConsignmentType),
+      arguments = Nil,
+      resolve = ctx => ctx.ctx.consignmentService.getConsignments(),
```
suzannehamilton, line 129: The change to the field definitions is simpler than I expected! 👍

src/test/resources/json/consignments_data_all.json
```diff
@@ -0,0 +1,54 @@
...
+        "createdDatetime": "2020-01-01T09:00:00Z",
+        "transferInitiatedDatetime": "2020-01-01T09:00:00Z",
+        "userid": "4ab14990-ed63-4615-8336-56fbb9960300",
+        "consignmentid": "c31b3d3e-1931-421b-a829-e2ef4cd8930c",
+        "consignmentReference": "TEST-TDR-2021-MTB",
+        "totalFiles": 0,
+        "files": [],
+        "fileChecks": {
+          "antivirusProgress": {
+            "filesProcessed": 0
```
suzannehamilton, line 16: Given that this is a test of retrieving all data, is it possible to add a few files with a bit of progress? Then we can check that the deferred resolvers are actually calculating something on this endpoint.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -56,6 +56,11 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
     db.run(query.result)
   }
 
+  def getConsignments(): Future[Seq[ConsignmentRow]] = {
+    val query = Consignment
+    db.run(query.result)
```
suzannehamilton, line 61: Sorry, I've just realised I'm probably looking at this PR too early. I'm guessing you'll be adding a sort order and cursor here?

TomJKing: Yes, I've not actually implemented the pagination yet.

PR #153, opened 2021-06-18
Tdr 1197 reporting consignments

TomJKing:


schema.graphql
```diff
@@ -93,6 +93,24 @@ type Consignment {
   currentStatus: CurrentStatus!
 }
 
+"A connection to a list of items."
```
suzannehamilton, line 96: Haven't seen these annotations before - are they generated by Relay?

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -182,3 +183,16 @@ object ValidateHasExportAccess extends SyncAuthorisationTag {
     }
   }
 }
+
+object ValidateHasReportingAccess extends SyncAuthorisationTag {
```
suzannehamilton, line 187: Will we have to add this permission to the `getConsignment` endpoint as well?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -114,16 +120,47 @@ object ConsignmentFields {
...
   val queryFields: List[Field[ConsignmentApiContext, Unit]] = fields[ConsignmentApiContext, Unit](
     Field("getConsignment", OptionType(ConsignmentType),
       arguments = ConsignmentIdArg :: Nil,
       resolve = ctx => ctx.ctx.consignmentService.getConsignment(ctx.arg(ConsignmentIdArg)),
       tags = List(ValidateUserHasAccessToConsignment(ConsignmentIdArg))
+    ),
+    Field("consignments", consignmentConnections,
+      arguments = List(LimitArg, CurrentCursorArg),
+      resolve = ctx => {
+        val limit: Int = ctx.args.arg("limit")
```
suzannehamilton, line 144: What's the maximum limit? How do we stop the client from requesting too much data at once?

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -56,6 +56,18 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
     db.run(query.result)
   }
 
+  def getConsignments(limit: Int, after: String): Future[Seq[ConsignmentRow]] = {
+    val criteriaAfter = Option(after)
+
+    val query = Consignment.filter { c =>
+      List(
+        criteriaAfter.map(c.consignmentreference >= _))
+        .collect({case Some(criteria) => criteria})
```
suzannehamilton, line 65: Why does it turn `after` into an `Option` and then ignore the None case?

suzannehamilton (commented):
The general approach looks good to me, thanks. I haven't finished reviewing yet - I'm just posting the comments I've got so far.

schema.graphql
TomJKing: Yes

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
TomJKing: If Phil will be using the `getConsignment` endpoint then I guess so.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: Good point. Forgot to add a maximum

src/test/resources/json/consignments_query_alldata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "query": "{ consignments(limit: 2, currentCursor: \"consignment-ref1\") { pageInfo { startCursor, endCursor, hasNextPage, hasPreviousPage }, edges { node { seriesid, consignmentid, userid, totalFiles, fileChecks { antivirusProgress { filesProcessed }, checksumProgress { filesProcessed }, ffidProgress { filesProcessed }}, parentFolder, series { code }, transferringBody { name, tdrCode }, consignmentReference, createdDatetime, transferInitiatedDatetime, exportDatetime files { fileId metadata { clientSideFileSize, clientSideLastModifiedDate, clientSideOriginalFilePath, foiExemptionCode, heldBy, language, legalStatus, rightsCopyright } ffidMetadata { software softwareVersion binarySignatureFileVersion containerSignatureFileVersion method matches { extension identificationBasis puid } datetime }} currentStatus { upload }}}}}"
```
suzannehamilton, line 2: I know a lot of our existing queries are like this, but is it valid to split the GraphQL query across multiple lines? In the past, I've found it quite hard to debug queries like this because you have to do so much horizontal scrolling.

src/test/resources/json/consignments_query_somedata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "query": "{ consignments(limit: 2, currentCursor: \"consignment-ref1\") { pageInfo { startCursor, endCursor, hasNextPage, hasPreviousPage }, edges { node { consignmentid, consignmentReference }}}}"
```
suzannehamilton, line 2: Out of curiosity, if you don't include the `pageInfo` in the query, does the response just skip it like regular GraphQL response properties?

src/test/resources/json/getconsignment_data_all.json
```diff
@@ -21,13 +21,13 @@
...
             "clientSideOriginalFilePath": "ClientSideOriginalFilepath value",
-            "legalStatus": "LegalStatus value",
+            "legalStatus": "Public Record",
             "clientSideFileSize": 1,
-            "clientSideLastModifiedDate": "2021-03-11T12:30:30.592853",
-            "rightsCopyright": "RightsCopyright value",
-            "heldBy": "HeldBy value",
-            "foiExemptionCode": "FoiExemptionCode value",
-            "language": "Language value"
+            "clientSideLastModifiedDate": "2021-02-08T16:00",
```
suzannehamilton, line 26: What caused the date format change?

src/test/resources/json/getconsignment_data_all.json
```diff
@@ -105,7 +105,7 @@
       "parentFolder": "ALL CONSIGNMENT DATA PARENT FOLDER",
       "fileChecks": {
         "antivirusProgress": {
-          "filesProcessed": 1
+          "filesProcessed": 2
```
suzannehamilton, line 108: Why has this changed? IIRC, it was deliberately checking a different value to `checksumProgress`.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -7,92 +7,90 @@ import org.scalatest.flatspec.AnyFlatSpec
...
 
+  val consignmentIdOne: UUID = UUID.fromString("20fe77a7-51b3-434c-b5f6-a14e814a2e05")
+  val consignmentIdTwo: UUID = UUID.fromString("fa19cd46-216f-497a-8c1d-6caaf3f421bc")
+  val consignmentIdThree: UUID = UUID.fromString("614d0cba-380f-4b09-a6e4-542413dd7f4a")
+  val consignmentIdFour: UUID = UUID.fromString("47019574-8407-40c7-b618-bf2b8f8b0de7")
+
   "addParentFolder" should "add parent folder name to an existing consignment row" in {
     val db = DbConnection.db
     val consignmentRepository = new ConsignmentRepository(db, new CurrentTimeSource)
-    val consignmentId = UUID.fromString("0292019d-d112-465b-b31e-72dfb4d1254d")
```
suzannehamilton, line 21: What's the reason for moving these out of the test? I'm a bit wary of shared test setup, because it can cause issues when you try to change the values for one test, and it breaks a bunch of others. (Or worse, they still pass but not for the original reason.)

It also makes the different values look significant, so a reader might start wondering why this test uses `consignmentIdThree` but this other one uses `consignmentIdFour`. What's special about the values? If they're created inside the test, it's obvious it's just a local value that doesn't depend on anything else.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -103,4 +101,73 @@ class ConsignmentRepositorySpec extends AnyFlatSpec with TestDatabase with Scala
...
+  "getConsignments" should "return no consignments where limit set at '0'" in {
+    val db = DbConnection.db
+    val consignmentRepository = new ConsignmentRepository(db, new CurrentTimeSource)
+    createConsignments()
+
+    val response = consignmentRepository.getConsignments(0, "TDR-2021-A").futureValue
+    response.size should be (0)
+  }
+
+  "getConsignments" should "return no consignments where non-existent cursor value provided'" in {
```
suzannehamilton, line 150: Just to check - is this the Relay convention?

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -103,4 +101,73 @@ class ConsignmentRepositorySpec extends AnyFlatSpec with TestDatabase with Scala
...
+    consignmentIds.contains(consignmentIdTwo) should be (true)
+    consignmentIds.contains(consignmentIdThree) should be (true)
+  }
+
+  "getConsignments" should "return all consignments up to limit where empty cursor provided" in {
+    val db = DbConnection.db
+    val consignmentRepository = new ConsignmentRepository(db, new CurrentTimeSource)
+    createConsignments()
+
+    val response = consignmentRepository.getConsignments(2, "").futureValue
```
suzannehamilton, line 134: As above, using an empty string is a bit hard to follow because the meaning is not obvious from the type system.

If the empty string _is_ the Relay convention, could we convert it into an `Option` in the service layer so the repository can deal with a cursor that's got a type of `Option[String]` that makes it more obvious that it is sometimes empty?

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -103,4 +101,73 @@ class ConsignmentRepositorySpec extends AnyFlatSpec with TestDatabase with Scala
...
+    response.size should be (2)
+    val consignmentIds: List[UUID] = response.map(cr => cr.consignmentid).toList
+    consignmentIds.contains(consignmentIdOne) should be (true)
```
suzannehamilton, line 137: What kind of message does this assertion give when it fails? Is it something opaque like `expected 'true' but got 'false'`?

ScalaTest matchers like `consignmentIds should contain (consignmentIdOne)` are normally better because the message is more helpful, e.g. `List('a', 'b') did not contain element 'c'`.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -296,7 +229,7 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
 
     val response: GraphqlQueryData = runTestQuery("query_somedata", validUserToken())
 
-    response.errors should have size 1
+    response.errors.size should be(1)
```
suzannehamilton, line 232: What's the reason for this change? I'd have thought the original would give a better failure message.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -207,4 +207,136 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
...
+    val consignmentRows: List[ConsignmentRow] = consignmentRowParams.map(p => createConsignmentRow(p._1, p._2, p._3))
+
+    val limit = 2
+
+    val mockResponse: Future[Seq[ConsignmentRow]] = Future.successful(consignmentRows)
+    when(consignmentRepoMock.getConsignments(limit + 1, "consignment-ref1")).thenReturn(mockResponse)
+
+    val response: PaginatedConsignments = consignmentService.getConsignments(limit, "consignment-ref1").futureValue
+
+    verify(consignmentRepoMock, times(1)).getConsignments(limit + 1, "consignment-ref1")
```
suzannehamilton, line 231: Minor thing, but you don't strictly need to call `verify` when your test is checking something without side-effects (in this case, fetching data from the repository). It's a repeat of the setup on line 227, and the interesting assertions are the ones below - does it return the right data?

suzannehamilton (commented):
I've been through all the code now. There are quite a few comments because there are a lot of changes, but the general approach still looks good. 👍

src/test/resources/json/consignments_query_somedata.json
TomJKing: Yes

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -54,6 +54,17 @@ class ConsignmentService(
       row => convertRowToConsignment(row)))
   }
 
+  def getConsignments(limit: Int, currentCursor: String): Future[PaginatedConsignments] = {
+    for {
+      //Limit increased by one to get the value of the next cursor
+      response <- consignmentRepository.getConsignments(limit + 1, currentCursor)
```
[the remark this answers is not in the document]
suzannehamilton: Did you say the Relay convention is to use the final cursor? Do you have a link? I'm just curious to see how the docs describe it.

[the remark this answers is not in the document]
TomJKing: Link: https://relay.dev/graphql/connections.htm#sec-Forward-pagination-arguments

> You should generally pass the cursor of the last edge in the previous page for after.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -54,6 +57,17 @@ class ConsignmentService(
       row => convertRowToConsignment(row)))
   }
 
+  def getConsignments(limit: Int, currentCursor: Option[String]): Future[PaginatedConsignments] = {
+    val l: Int = if (limit > maxLimit) maxLimit else limit
+
+    for {
+      response <- consignmentRepository.getConsignments(l, currentCursor)
```
suzannehamilton, line 64: Could `l` be `maxConsignments` to make this easier to understand?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -346,11 +404,47 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
     result.getString(field)
   }
 
-  private def createSeries(bodyId: UUID): Unit = {
-    val sql = "INSERT INTO Series (SeriesId, BodyId) VALUES (?,?)"
-    val ps: PreparedStatement = databaseConnection.prepareStatement(sql)
-    ps.setString(1, "6e3b76c4-1745-4467-8ac5-b4dd736e1b3e")
-    ps.setString(2, bodyId.toString)
-    ps.executeUpdate()
+  private def setUpConsignments(consignmentParams: List[(UUID, String, List[UUID])]): Unit = {
```
suzannehamilton, line 407: Might be worth making a case class? It's easier to follow than a tuple because the parameters have names rather than just positions.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -207,4 +211,182 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
...
+    )
+
+    val consignmentRows: List[ConsignmentRow] = consignmentRowParams.map(p => createConsignmentRow(p._1, p._2, p._3))
+
+    val limit = 3
+
+    val mockResponse: Future[Seq[ConsignmentRow]] = Future.successful(consignmentRows)
+    when(consignmentRepoMock.getConsignments(2, Some("consignment-ref1"))).thenReturn(mockResponse)
+
+    val response: PaginatedConsignments = consignmentService.getConsignments(2, Some("consignment-ref1")).futureValue
```
suzannehamilton, line 260: This looks like it doesn't use the limit of 3?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -207,4 +211,182 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
...
+    val mockResponse: Future[Seq[ConsignmentRow]] = Future.successful(Seq())
+    when(consignmentRepoMock.getConsignments(limit, Some("consignment-ref1"))).thenReturn(mockResponse)
+
+    val response: PaginatedConsignments = consignmentService.getConsignments(limit, Some("consignment-ref1")).futureValue
+
+    response.lastCursor should be (None)
+    response.consignmentEdges should have size 0
+  }
+
+  "convertToEdges" should "return consignment edges for the given consignment rows" in {
```
suzannehamilton, line 356: If this is a private method, then it's being implicitly tested by the other tests. You shouldn't need to test methods which aren't part of a class's public interface, partly because it makes it harder to refactor things later.

suzannehamilton (commented):
Just a few small comments, thanks!

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
TomJKing: Phil still wants to use the `getConsignment` endpoint

TomJKing: Spoke to Phil. Will go with just access to the `consignments` endpoint for now. Will add access to `getConsignment` if it appears he will need it later.

src/test/resources/json/consignments_query_alldata.json
TomJKing: I'm not sure how straight forward this is, as json doesn't support multi-line strings.

suzannehamilton (approved):
Looks good, thanks. 🎉

PR #155, opened 2021-06-30
Log all exceptions at the Akka route layer

MancunianSam:
- Rethrow any exceptions.
- Add an exception handler.
- Add a test for the exception handler behaviour.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/ErrorCodes.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/ErrorCodes.scala
@@ -10,4 +10,5 @@ object ErrorCodes {
   val notAuthorised = "NOT_AUTHORISED"
   val invalidConsignmentState = "INVALID_CONSIGNMENT_STATE"
   val invalidInputData = "INVALID_INPUT_DATA"
+  val unknownError = "UNKNOWN_ERROR"
 }
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
@@ -2,21 +2,25 @@ package uk.gov.nationalarchives.tdr.api.http
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import akka.http.scaladsl.model.StatusCodes.InternalServerError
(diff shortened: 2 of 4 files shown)


TomJKing (dismissed):

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/ErrorCodes.scala
```diff
@@ -10,4 +10,5 @@ object ErrorCodes {
   val notAuthorised = "NOT_AUTHORISED"
   val invalidConsignmentState = "INVALID_CONSIGNMENT_STATE"
   val invalidInputData = "INVALID_INPUT_DATA"
+  val unknownError = "UNKNOWN_ERROR"
```
suzannehamilton, line 13: This doesn't look like it's used any more.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
```diff
@@ -30,6 +34,33 @@ class Routes(val config: Config) extends Cors {
...
+    ExceptionHandler {
+      case ex: Throwable =>
+        entity(as[JsValue]) { requestJson =>
+          val JsObject(fields) = requestJson
+          val JsString(query) = fields("query")
+
+          QueryParser.parse(query) match {
+            case Failure(parseException) =>
+              logger.error(s"Error parsing entity $requestJson", parseException)
+              complete(HttpResponse(InternalServerError, entity = parseException.getMessage))
```
suzannehamilton, line 47: It looks like we're returning the error about the parse exception, but we're not including any information about the original exception. Am I following that right? Would it be better to return an internal server error based on the original error, since the parsing was just done to get some more context?

suzannehamilton (changes requested):

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/ErrorCodes.scala
MancunianSam: Good spot, I've removed it.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Routes.scala
MancunianSam: Yeah we're returning the parse exception but it would make more sense to have the original exception. I've left the log line for the parse exception because it might be useful. I've added in a log line for the original exception which just includes the message and returned the exception message in the response.

suzannehamilton (approved):

PR #156, opened 2021-07-08
Create start upload mutation

MancunianSam:
- Add startUpload mutation.
- Add a test for the new service method

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
@@ -32,7 +32,7 @@ class ConsignmentService(
     consignmentStatusRepository.getConsignmentStatus(startUploadInput.consignmentId).flatMap(status => {
       val uploadStatus = status.find(s => s.statustype == "Upload")
       if(uploadStatus.isDefined) {
-        throw ConsignmentStateException(s"Consignment status is ${uploadStatus.get.value}")
+        throw ConsignmentStateException(s"Existing consignment upload status is '${uploadStatus.get.value}', so cannot start new upload")
       }
       val now = Timestamp.from(timeSource.now)
       val consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, startUploadInput.consignmentId, "Upload", "InProgress", now)
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
@@ -17,7 +17,6 @@ import scala.math.min
 
 class ConsignmentService(
                           consignmentRepository: ConsignmentRepository,
-                          consignmentStatusRepository: ConsignmentStatusRepository,
                           fileMetadataRepository: FileMetadataRepository,
                           fileRepository: FileRepository,
                           ffidMetadataRepository: FFIDMetadataRepository,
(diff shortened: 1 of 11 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -27,6 +28,15 @@ class ConsignmentService(
 
   val maxLimit: Int = config.getInt("pagination.consignmentsMaxLimit")
 
+  def startUpload(startUploadInput: StartUploadInput): Future[Int] = {
+    val now = Timestamp.from(timeSource.now)
+    val consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, startUploadInput.consignmentId, "Upload", "InProgress", now)
+    for {
+      _ <- consignmentRepository.addParentFolder(startUploadInput.consignmentId, startUploadInput.parentFolder)
+      res <- consignmentStatusRepository.addConsignmentStatus(consignmentStatusRow)
```
suzannehamilton, line 36: How hard would it be to put these into the same transaction? The old status update happened transactionally: https://github.com/nationalarchives/tdr-consignment-api/blob/bc8b36aedb4337d2359964b2edc12fc26307bb15/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala#L23

MancunianSam: Yeah makes sense. It's slightly fewer changes because you don't have to pass the `consignmentMetadataRepository` into the `ConsignmentService`. I've added another `addParentFolder` method into `ConsignmentRepository` so it doesn't break the code already calling that. Once we remove the old `addFiles` field, I'll get rid of the old method as well.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -394,6 +399,16 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
     response.errors.head.extensions.get.code should equal("NOT_AUTHORISED")
   }
 
+  "startUpload" should "add the upload status and update the parent folder" in {
```
suzannehamilton, line 402: Should have noticed this earlier, sorry, but what about the case where `startUpload` is called twice? The second one should be rejected, like a second `createFiles` is now.

MancunianSam: Good point, I've set it to reject if the consignment status for "Upload" is "InProgress" or "Complete". I've added a couple of tests to check for it.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -27,6 +28,19 @@ class ConsignmentService(
 
   val maxLimit: Int = config.getInt("pagination.consignmentsMaxLimit")
 
+  def startUpload(startUploadInput: StartUploadInput): Future[String] = {
+    consignmentStatusRepository.getConsignmentStatus(startUploadInput.consignmentId).flatMap(status => {
+      val uploadStatus = status.find(s => s.statustype == "Upload")
+      if(uploadStatus.isDefined) {
+        throw ConsignmentStateException(s"Consignment status is ${uploadStatus.get.value}")
```
suzannehamilton, line 35: I'm trying to think of what it would be useful to see in the error if it actually happened. What about `s"Existing consignment upload status is '${uploadStatus.get.value}', so cannot start new upload"`?

MancunianSam: That is more descriptive. I've changed it here and updated the tests.

suzannehamilton (approved):

PR #158, opened 2021-07-20
Add endpoint for checking db health

techncl:


src/main/scala/uk/gov/nationalarchives/tdr/api/http/ApiServer.scala
```diff
@@ -21,7 +22,7 @@ object ApiServer extends App {
 
   scala.sys.addShutdownHook(() -> shutdown())
 
-  val routes = new Routes(ConfigFactory.load())
+  val routes = new Routes(ConfigFactory.load(), DbConnection.db)
```
MancunianSam, line 25: This argument in the `Routes` class isn't being used so you can take it out.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FullHealthCheckService.scala
```diff
@@ -0,0 +1,18 @@
...
+import slick.jdbc.PostgresProfile.api._
+import uk.gov.nationalarchives.tdr.api.db.DbConnection
+import uk.gov.nationalarchives.tdr.api.db.repository.TransferringBodyRepository
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class FullHealthCheckService {
+
+  def checkDbIsUpAndRunning(db: Database): Future[Unit] = {
```
MancunianSam, line 12: Do you need to pass the database in here? We're not passing the database in any other methods I think so it might be better to either create the `TransferringBodyRepository` directly in the class like
`val transferringBodyRepository = new TransferringBodyRepository(DbConnection.db)`

Or pass the repository into the service like we do with the others so the class declaration would  be `class FullHealthCheckService(transferringBodyRepository: TransferringBodyRepository)`

I would be tempted to do the first one so we don't have to pass that `Database` object round a lot but it's up to you.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/TransferringBodyRepositorySpec.scala
```diff
@@ -0,0 +1,36 @@
...
+import uk.gov.nationalarchives.tdr.api.utils.TestUtils._
+
+import java.util.UUID
+import scala.concurrent.ExecutionContext
+
+
+class TransferringBodyRepositorySpec extends AnyFlatSpec with TestDatabase with ScalaFutures with Matchers {
+  implicit val executionContext: ExecutionContext = ExecutionContext.Implicits.global
+
+  "dbHasTransferringBodies" should "Should return true if db has 1 Transferring Bodies in it" in {
```
MancunianSam, line 17: You've got an extra "Should" in here and in the test below.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/CorsSpec.scala
```diff
@@ -18,7 +19,7 @@ class CorsSpec extends AnyFlatSpec with Matchers with ScalatestRouteTest {
 
   private val config = ConfigFactory.load()
     .withValue("frontend.urls", ConfigValueFactory.fromIterable(crossOriginUrls))
-  private val route = new Routes(config).route
+  private val route = new Routes(config, DbConnection.db).route
```
MancunianSam, line 23: Same as above but you don't need to pass this into the Routes class.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/RouteAuthenticationSpec.scala
```diff
@@ -39,4 +50,49 @@ class RouteAuthenticationSpec extends AnyFlatSpec with Matchers with ScalatestRo
...
+                db
+            }
+          case _ =>
+            db
+        }
+      }
+    }
+
+    val testDb = DbConnectionTest.db
+    val route = new Routes(ConfigFactory.load(), testDb).route
```
MancunianSam, line 90: Because this parameter is not used in the `Routes` class, this is testing the same thing as the test above. I think checking that we get an error if the service throws an exception when the table is empty is enough and you can get rid of this one.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/TransferringBodyRepository.scala
```diff
@@ -16,4 +16,9 @@ class TransferringBodyRepository(db: Database) {
 
     db.run(query.result).map(body => body.head)
   }
+
+  def dbHasTransferringBodies(implicit executionContext: ExecutionContext): Future[Boolean] = {
+    val query = Body.length
```
MancunianSam, line 21: I know the other method in here has an execution context on the method. Now there's two methods with it, could you put it on the class level and remove it from the method level.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FullHealthCheckService.scala
```diff
@@ -0,0 +1,18 @@
+package uk.gov.nationalarchives.tdr.api.service
+
+import slick.jdbc.PostgresProfile.api._
+import uk.gov.nationalarchives.tdr.api.db.DbConnection
+import uk.gov.nationalarchives.tdr.api.db.repository.TransferringBodyRepository
+
+import scala.concurrent.ExecutionContext.Implicits.global
```
MancunianSam, line 7: Instead of using the global execution context, can you pass it into the service constructor.

MancunianSam (commented):
Looks good, cheers. Just a few comments about some arguments to the classes

src/main/scala/uk/gov/nationalarchives/tdr/api/http/ApiServer.scala
techncl: Ah, thanks. That was a mistake, the Routes class is supposed to use the db passed into it, but what is weird is that the `"The db" should "return 500 Internal Server Error if the db is down"` test was still working even though it closes the db and now works when I make the Routes class use the db that has been passed to it

techncl: Ahhh, I know why. It's because that endpoint throws the same error type for two reasons: if there are no bodies in the DB AND if the db is down so it was just catching the former as it was essentially the same as the test above it (since it wasn't using the new closed test db that I added) 🤦‍♂️ . Good catch; thanks 👍

MancunianSam (approved):

PR #159, opened 2021-07-21
Replace addFiles and addClientFileMetadata

MancunianSam:
These have been replaced with startUpload and addFilesAndMetadata. The
front end and the end to end tests have been updated to use the new mutations
so we can remove these now.


src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
```diff
@@ -92,29 +91,6 @@ case class ValidateUserHasAccessToConsignment[T](argument: Argument[T]) extends
   }
 }
 
-object ValidateUserOwnsFiles extends AuthorisationTag {
```
TomJKing, line 95: Assume a different AuthorisationTag is used for the new endpoints, and this is no longer needed as was only used by the old endpoints that are being deleted?

TomJKing (commented):
Looks good, just one minor clarification

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
MancunianSam: Sort of. Before, we were creating the files and adding the metadata in a separate API call so we could check that the user owned the files but now we're creating the files and metadata in the same call so we can't and don't need to check any more. We're still checking the user owns the consignment on the new endpoint.

TomJKing (approved):

PR #162, opened 2021-08-05
Ensure only files with ffid matches are counted

TomJKing:
There is potential that the ffid matches will not contain any results

In this case ensure that the file checks do not complete


MancunianSam (approved):
Looks good, thanks.

PR #163, opened 2021-08-05
Do not add ffid metadata if no ffid matches

TomJKing:
There is the potential that no ffid matches are returned

When this happens no ffid metadata should be added to prevent invalid data being stored in the database


suzannehamilton (approved):
This looks good, thanks 👍

PR #167, opened 2021-08-20
Remove upload complete status from add file

techncl:


suzannehamilton (approved):
Makes sense as the first step to removing the field from this endpoint. 👍

PR #168, opened 2021-09-06
Remove 'isComplete' argument from file fields as it's not needed

techncl:


suzannehamilton:
Is this safe to merge? Has the frontend change that removes it been deployed to production?

techncl:
Ah yes! No, they haven't. I will close this until then.

suzannehamilton (approved):

PR #169, opened 2021-09-08
Add Transfer Agreement status flag

techncl:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementService.scala
```diff
@@ -1,23 +1,32 @@
...
+class TransferAgreementService(consignmentMetadataRepository: ConsignmentMetadataRepository,
+                               consignmentStatusRepository: ConsignmentStatusRepository,
+                               uuidSource: UUIDSource, timeSource: TimeSource) (implicit val executionContext: ExecutionContext) {
 
   def addTransferAgreement(input: AddTransferAgreementInput, userId: UUID): Future[TransferAgreement] = {
-    consignmentMetadataRepository.addConsignmentMetadata(convertInputToPropertyRows(input, userId)).map(
+    val transferAgreement = consignmentMetadataRepository.addConsignmentMetadata(convertInputToPropertyRows(input, userId)).map(
       rows => convertDbRowsToTransferAgreement(input.consignmentId, rows)
     )
+    addTransferAgreementStatus(input.consignmentId, "Completed")
```
MancunianSam, line 23: The result of this method call isn't assigned to anything. Because it's a Future, it's eager and it's still executed but because you're not mapping or flat mapping, you might have problems getting any errors back to the client and you might have problems if the database is slow and the Future doesn't complete before Akka returns and kills the thread.

I like the for comprehension so you can do it like this:
```scala
for {
      transferAgreement <- consignmentMetadataRepository.addConsignmentMetadata(convertInputToPropertyRows(input, userId)).map(
        rows => convertDbRowsToTransferAgreement(input.consignmentId, rows)
      )
      _ <- addTransferAgreementStatus(input.consignmentId, "Completed")
    } yield transferAgreement
```
If you don't need the for comprehension then you can use map and flatMap
```scala
    consignmentMetadataRepository.addConsignmentMetadata(convertInputToPropertyRows(input, userId)).map(
      rows => convertDbRowsToTransferAgreement(input.consignmentId, rows)
    ).flatMap(transferAgreement => {
      addTransferAgreementStatus(input.consignmentId, "Completed").map(_ => transferAgreement)
    })
```
For either one, you're waiting for the Future to complete or fail before the `addTransferAgreement` method returns.
This is the main reason that Future is a pain to work with :smile:

techncl: Ah, of course; this makes sense. Thanks, I'll add the for comprehension

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentStatusRepositorySpec.scala
```diff
@@ -4,16 +4,41 @@ import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import uk.gov.nationalarchives.tdr.api.db.DbConnection
-import uk.gov.nationalarchives.tdr.api.utils.{TestDatabase, TestUtils}
+import uk.gov.nationalarchives.tdr.api.utils.TestDatabase
+import uk.gov.nationalarchives.tdr.api.utils.TestUtils
+import uk.gov.nationalarchives.Tables.ConsignmentstatusRow
 
 import java.sql.Timestamp
+import java.time.Instant
```
MancunianSam, line 12: This import is unused

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusServiceSpec.scala
```diff
@@ -47,7 +47,7 @@ class ConsignmentStatusServiceSpec extends AnyFlatSpec with MockitoSugar with Re
 
     val response = consignmentService.getConsignmentStatus(consignmentId).futureValue
 
-    response should be(CurrentStatus(None))
+    response should be(CurrentStatus(None, None))
```
MancunianSam, line 50: This is just a suggestion so feel free to ignore me but do you think it's worth adding another couple of tests in here. One to check that if you only return the `transferAgreement` status then the `upload` status is `None` and vice versa.  What do you think?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementServiceSpec.scala
```diff
@@ -2,32 +2,35 @@ package uk.gov.nationalarchives.tdr.api.service
 
 import java.sql.Timestamp
 import java.util.UUID
-
 import org.mockito.ArgumentMatchers.any
-import org.mockito.MockitoSugar
+import org.mockito.{ArgumentCaptor, MockitoSugar}
```
MancunianSam, line 6: The ArgumentCaptor import is unused.

MancunianSam (commented):
Looks good. Apart from the Future map thing, there's just a couple of unused imports and a question about adding a couple more tests

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusServiceSpec.scala
techncl: Yes, that's a good idea; I'll add these.

MancunianSam (approved):

PR #171, opened 2021-10-06
Remove unnecessary isAgreementComplete check and field

techncl:


TomJKing (commented):
Looks good. Have you checked that the e2e tests don't need adjusting following this change?

PR #175, opened 2021-11-03
Add 'consignment type'

TomJKing:


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -114,12 +115,49 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
     checkConsignmentExists(response.data.get.addConsignment.consignmentid.get)
   }
 
+  "addConsignment" should "create a consignment of type 'standard' if no consignment type provided" in {
+    addSeries(
+      UUID.fromString("6e3b76c4-1745-4467-8ac5-b4dd736e1b3e"),
+      bodyId = UUID.fromString("830f0315-e683-440e-90d0-5f4aa60388c6"),
```
MancunianSam, line 121: Is it worth moving the two `UUID.fromString` calls into a variable at the top of the class and referencing it here as this is called quite a few  times?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -114,12 +115,49 @@ class ConsignmentRouteSpec extends AnyFlatSpec with Matchers with TestRequest wi
...
+      "Mock series")
+
+    val expectedResponse: GraphqlMutationData = expectedMutationResponse("data_judgment_consignment_type")
+    val response: GraphqlMutationData = runTestMutation("mutation_judgment_consignment_type", validUserToken(body = "default-transferring-body-code"))
+    response.data.get.addConsignment should equal(expectedResponse.data.get.addConsignment)
+
+    checkConsignmentExists(response.data.get.addConsignment.consignmentid.get)
+  }
+
```
MancunianSam, line 143: Is it worth adding a test in to check that a standard consignment is created when a standard consignment type is provided as well as when it's missing?

MancunianSam (commented):
Looks good cheers. Just a couple of suggestions

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
TomJKing: Just moved the two UUID creation calls into variables. It gets too confusing to do much else. It's better to see the workings clearly of each individual test, rather than abstract the tests too much, I think.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
TomJKing: This was already done in the first `addConsignment` test. So just renamed the test to make it clearer

MancunianSam (approved):

PR #176, opened 2021-11-05
Fix test failing due to time difference between linux and windows

thanhz:
image
Fix some failing tests due to rounding of time on windows machines


MancunianSam (approved):
Thanks for sorting this.

PR #178, opened 2021-11-11
Add Database endpoint and Graphql Schema for Final Judgment Transfers

thanhz:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FinalTransferConfirmationService.scala
```diff
@@ -51,3 +73,11 @@ object FinalTransferConfirmationService {
     LegalOwnershipTransferConfirmed
   )
 }
+
+object FinalJudgmentTransferConfirmationService {
```
MancunianSam, line 77: You don't need a second object here. It does work but the existing `FinalTransferConfirmationService` object on line 67 is the companion object for the class so you can add in `LegalCustodyTransferConfirmed` and `finalJudgmentTransferConfirmationProperties` into the existing object.

src/test/resources/json/addfinaljudgmenttransferconfirmation_mutation_somedata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "query": "mutation {addFinalJudgmentTransferConfirmation(addFinalJudgmentTransferConfirmationInput: {consignmentId: \"b42dccf0-549a-4204-bc9e-c6b69560b7a5\", legalCustodyTransferConfirmed: true}) {consignmentId, legalCustodyTransferConfirmed}}"
+}
```
MancunianSam, line 3: Could you put a line break at the end of the file please, keeps things consistent.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FinalTransferConfirmationRouteSpec.scala
```diff
@@ -43,6 +60,16 @@ class FinalTransferConfirmationRouteSpec extends AnyFlatSpec with Matchers with
     checkFinalTransferConfirmationExists(consignmentId)
   }
 
+  it should "return all requested fields from inserted final judgment transfer confirmation consignment metadata properties" in {
```
MancunianSam, line 63: Could you change `it` to `"The api` It does work as you've got it but all the other tests are in the `"" should ""` format.

MancunianSam (commented):
Looks good thanks. Just a couple of small things.

MancunianSam (approved):

PR #184, opened 2021-12-06
Add folder information

MancunianSam:


src/main/scala/uk/gov/nationalarchives/tdr/api/model/file/FileType.scala
```diff
@@ -0,0 +1,38 @@
+package uk.gov.nationalarchives.tdr.api.model.file
+
+import uk.gov.nationalarchives.tdr.api.graphql.DataExceptions.InputDataException
+
+object FileType {
```
techncl, line 5: This is probably too late since we have a column with this name, but maybe the name could be something like "ItemType", "ObjectType", "ResourceType" or "ContentType"? "FileType" sounds like the type of file e.g. pdf, png, mp3 etc.

MancunianSam: I've changed the object name to NodeType as per the slack discussion.

techncl (commented):
Looks good! Just had a question about generating the `FileRows`

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/TreeNodesUtils.scala
```diff
@@ -0,0 +1,36 @@
...
+import uk.gov.nationalarchives.tdr.api.utils.TreeNodesUtils.TreeNode
+
+import scala.annotation.tailrec
+
+class TreeNodesUtils(uuidSource: UUIDSource) {
+  def generateNodes(filePaths: Set[String]): Map[String, TreeNode] = {
+    @tailrec
+    def innerFunction(originalPath: String, fileType: String, nodes: Map[String, TreeNode]): Map[String, TreeNode] = {
+      val jioFile = new JIOFile(originalPath)
+      val parentPath = Option(jioFile.getParent)
```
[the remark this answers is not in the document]
TomJKing: You're correct that having a leading `/` in the file path causes an entry without a `Name`. However, I don't think the client side files paths would include a leading `/`. That said I wondering whether we should be cleaning the file path string to remove leading `/` to prevent this from happening.

src/test/scala/uk/gov/nationalarchives/tdr/api/utils/TreeNodeUtilsSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.nationalarchives.tdr.api.model.file.NodeType
+
+class TreeNodeUtilsSpec extends AnyFlatSpec with MockitoSugar with Matchers with ScalaFutures {
+
+  "generateNodes" should "generate the correct nodes for a nested path" in {
+    val result = TreeNodesUtils(new FixedUUIDSource).generateNodes(Set("/a/path/with/some/nested/folders/file"))
+    val expectedSize = 8;
+    result.size should equal(expectedSize)
+    List(
+      ("/", None, ""),
```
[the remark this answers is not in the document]
TomJKing: I would say no. I don't think file systems allow a folder with no name, though this might be one for David C to answer, just to be sure.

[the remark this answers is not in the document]
MancunianSam: Empty folder names don't make much sense I don't think. If I do what Tom suggests in the comment above and remove any leading slashes, it gets rid of the empty folder name and the parent folder has a name again. I've updated the tests to show this. Does that look better?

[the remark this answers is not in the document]
MancunianSam: I've spoken to Phil as well and he's happy with the parent folder being the name of the folder after the first slash.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/TreeNodesUtils.scala
```diff
@@ -0,0 +1,37 @@
...
+class TreeNodesUtils(uuidSource: UUIDSource) {
+  def generateNodes(filePaths: Set[String]): Map[String, TreeNode] = {
+    @tailrec
+    def innerFunction(originalPath: String, fileType: String, nodes: Map[String, TreeNode]): Map[String, TreeNode] = {
+      val pathWithoutInitialSlash: String = if (originalPath.startsWith("/")) originalPath.tail else originalPath
+      val jioFile = new JIOFile(pathWithoutInitialSlash)
+      val parentPath = Option(jioFile.getParent)
+      val name = jioFile.getName
+      val treeNode = TreeNode(uuidSource.uuid, name, parentPath, fileType)
+      val nextMap = nodes + (pathWithoutInitialSlash -> treeNode)
```
techncl, line 20: Maybe this could just be done on line 29 instead of doing it each time `innerFunction` is called?

MancunianSam: I'm not sure what you mean. `nextMap` is recreated each time inner function is called with the current path and then returned at the end, I don't know how it can be moved out of `innerFunction`

techncl: Oops, sorry; I meant to make this comment on line 15.

MancunianSam: Yeah, you're right, it only needs to be done once. I've moved it out of the `innerFunction` thanks.

techncl (approved):

PR #188, opened 2022-01-28
Fix add transfer agreement compliance bug

techncl:


src/test/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementServiceSpec.scala
```diff
@@ -61,18 +60,10 @@ class TransferAgreementServiceSpec extends AnyFlatSpec with MockitoSugar with Ma
...
     transferAgreementResult.allEnglish shouldBe true
     transferAgreementResult.allPublicRecords shouldBe true
-
-    consignmentStatusResult.head.consignmentid shouldBe consignmentId
-    consignmentStatusResult.head.consignmentstatusid shouldBe consignmentStatusId
-    consignmentStatusResult.head.statustype shouldBe statusType
-    consignmentStatusResult.head.value shouldBe statusValue
   }
 
   "addTransferAgreementCompliance" should "add the correct metadata given correct arguments and set TA status to Completed" in {
```
TomJKing, line 69: Not sure this test is actually checking that the status has been set to `Completed`

src/test/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementServiceSpec.scala
```diff
@@ -84,22 +75,21 @@ class TransferAgreementServiceSpec extends AnyFlatSpec with MockitoSugar with Ma
...
       row("AppraisalSelectionSignOffConfirmed", "true"),
       row("InitialOpenRecordsConfirmed", "true"),
       row("SensitivityReviewSignOffConfirmed", "true")
     ))
-    val statusType = "TransferAgreement"
-    val statusValue = "Completed"
 
-    val mockTaConsignmentStatus = ConsignmentstatusRow(consignmentStatusId, consignmentId, statusType, statusValue, dateTime, None)
-    val mockTaConsignmentStatusResponse = Future.successful(Seq(mockTaConsignmentStatus))
+    val statusValue = "Completed"
```
TomJKing, line 89: This val doesn't appear to be used anywhere in the test?

src/test/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementServiceSpec.scala
```diff
@@ -131,19 +114,37 @@ class TransferAgreementServiceSpec extends AnyFlatSpec with MockitoSugar with Ma
...
+    val result: ConsignmentstatusRow = service.addTransferAgreementStatus(consignmentId).futureValue
 
     result.consignmentstatusid shouldBe consignmentStatusId
     result.consignmentid shouldBe consignmentId
     result.statustype shouldBe statusType
     result.value shouldBe statusValue
     result.createddatetime shouldBe createdTimestamp
   }
+
+  "updateTransferAgreementStatus" should "add the correct Transfer Agreement Status" in {
```
TomJKing, line 133: Not sure this test is actually checking that the correct status has been added. Maybe use `ArgumentCaptor`s to check that the correct arguments are being sent to the mocks. See for example https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/test/scala/uk/gov/nationalarchives/tdr/api/service/AntivirusMetadataServiceSpec.scala#L41 for example of using captors.

TomJKing (commented):
Looks good. Added some comments, mostly around the unit tests.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementService.scala
```diff
@@ -74,15 +74,19 @@ class TransferAgreementService(consignmentMetadataRepository: ConsignmentMetadat
       transferAgreementCompliance <- consignmentMetadataRepository.addConsignmentMetadata(convertTAComplianceInputToPropertyRows(input, userId)).map(
         rows => convertDbRowsToTransferAgreementCompliance(input.consignmentId, rows)
       )
-      _ <- addTransferAgreementStatus(input.consignmentId, "Completed")
+      _ <- updateTransferAgreementStatus(input.consignmentId, "Completed")
     } yield transferAgreementCompliance
   }
 
-  def addTransferAgreementStatus(consignmentId: UUID, statusValue: String): Future[ConsignmentstatusRow] = {
+  def addTransferAgreementStatus(consignmentId: UUID, statusValue: String="InProgress"): Future[ConsignmentstatusRow] = {
```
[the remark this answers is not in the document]
techncl: Yes, this is temporary, my plan is (Once we've deleted the old `addTransferAgreement` method on line 23), to remove the default parameter add `InProgress` as an argument to. `ConsignmentstatusRow` on line 82, i.e. replace `statusValue` (on line 82) with `InProgress` because I came to the conclusion that when you add a TA status, the only status it could/would/should be is an `InProgress` status (as that's the initial status value); everything else will be added via the `updateExistingTransferAgreementStatus`

src/test/scala/uk/gov/nationalarchives/tdr/api/service/TransferAgreementServiceSpec.scala
techncl: Ah, yes, you are correct, I was meant to remove this, but now that you mentioned the `ArgumentCaptor`, I might use that instead.

TomJKing (approved):
Looks good. Maybe worth putting in a separate JIRA task about reverting the default value InProgress", just so it doesn't get forgotten.

techncl:
> Looks good. Maybe worth putting in a separate JIRA task about reverting the default value InProgress", just so it doesn't get forgotten.

Thanks. Yes, I added one to Jira yesterday.

PR #189, opened 2022-02-04
Closure metadata API

thanhz:
Will update consignment-api-db dependency in build.sbt after [this](https://github.com/nationalarchives/tdr-consignment-api-data/pull/119) is merged

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationAuthoriser.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationAuthoriser.scala
@@ -21,14 +21,6 @@ class ValidationAuthoriser(implicit executionContext: ExecutionContext)
                            ctx: Context[ConsignmentApiContext, _]
                           ): BeforeFieldResult[ConsignmentApiContext, Unit] = {
 
-    // All fields must have an authorisation tag defined. This means that if we forget to add authorisation, the
-    // query is blocked by default, which prevents some security bugs.
-    val isTopLevelField = ctx.path.path.length == 1
-    val fieldHasAuthTag = ctx.field.tags.exists(tag => tag.isInstanceOf[AuthorisationTag])
-    if (isTopLevelField && !fieldHasAuthTag) {
-      throw new AssertionError(s"Query '${ctx.field.name}' does not have any authorisation steps defined")
-    }
-
     val validationList: Seq[BeforeFieldResult[ConsignmentApiContext, Unit]] = ctx.field.tags.map {
       case v: AuthorisationTag => {
          v.validate(ctx)
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
@@ -41,6 +41,21 @@ class FileMetadataRepository(db: Database)(implicit val executionContext: Execut
     db.run(query.result)
   }
 
+  def getClosureMetadataProperty(): Future[Seq[Filepropertyv2Row]] = {
+    val query = Filepropertyv2
+    db.run(query.result)
+  }
+
+  def getClosureMetadataValues(): Future[Seq[Filepropertyvaluesv2Row]] = {
+    val query = Filepropertyvaluesv2
+    db.run(query.result)
+  }
+
+  def getClosureMetadataDependencies(): Future[Seq[Filepropertydependanciesv2Row]] = {
+    val query = Filepropertydependanciesv2
+    db.run(query.result)
+  }
+
   def countProcessedChecksumInConsignment(consignmentId: UUID): Future[Int] = {
     val query = Filemetadata.join(File)
       .on(_.fileid === _.fileid).join(Fileproperty)
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
@@ -41,18 +41,18 @@ class FileMetadataRepository(db: Database)(implicit val executionContext: Execut
     db.run(query.result)
   }
 
-  def getClosureMetadataProperty(): Future[Seq[Filepropertyv2Row]] = {
+  def getClosureMetadataProperty: Future[Seq[Filepropertyv2Row]] = {
     val query = Filepropertyv2
     db.run(query.result)
   }
 
-  def getClosureMetadataValues(): Future[Seq[Filepropertyvaluesv2Row]] = {
+  def getClosureMetadataValues: Future[Seq[Filepropertyvaluesv2Row]] = {
     val query = Filepropertyvaluesv2
     db.run(query.result)
   }
 
-  def getClosureMetadataDependencies(): Future[Seq[Filepropertydependanciesv2Row]] = {
-    val query = Filepropertydependanciesv2
+  def getClosureMetadataDependencies: Future[Seq[Filepropertydependenciesv2Row]] = {
+    val query = Filepropertydependenciesv2
     db.run(query.result)
   }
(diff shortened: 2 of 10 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/MetadataFields.scala
```diff
@@ -0,0 +1,42 @@
+package uk.gov.nationalarchives.tdr.api.graphql.fields
```
TomJKing, line 1: I would be tempted to rename this: `CustomMetadataFields`, just to make it clearer what it is. The word `metadata` is used a lot in the code, and its going to be important to qualify what sort of metadata it is, to avoid confusion.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
```diff
@@ -41,6 +41,21 @@ class FileMetadataRepository(db: Database)(implicit val executionContext: Execut
     db.run(query.result)
   }
 
+  def getClosureMetadataProperty: Future[Seq[FilepropertyRow]] = {
```
TomJKing, line 44: I would be tempted to move these queries out into a new repo class: `CustomMetadataPropertiesRepository`. This will make it clearer what these are.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -24,6 +23,21 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
 
   val loggingUtils: LoggingUtils = LoggingUtils(Logger("FileMetadataService"))
 
+  def getPropertyType(propertyType: Option[String]): PropertyType = propertyType match {
```
TomJKing, line 26: I would be tempted to move these functions out into their own service: `CustomMetadataPropertiesService` just to make it clear what they are, and distinguish them from the actual "metadata" for files.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -82,6 +96,52 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
           )
       }
   }
+
+  def getClosureMetadata: Future[Seq[MetadataField]] = {
```
TomJKing, line 100: As the comment about, maybe move this function into their own service: `CustomMetadataPropertiesService`

src/test/resources/scripts/init.sql
```diff
@@ -139,3 +147,21 @@ CREATE TABLE IF NOT EXISTS "FileStatus"
...
+
+CREATE TABLE IF NOT EXISTS "FilePropertyValues"
+(
+    "PropertyName" varchar(255) NOT NULL,
+    "PropertyValue" varchar(255) NOT NULL,
+    "Default" boolean NULL,
+    "Dependencies" integer NULL,
+    "SecondaryValue" integer NULL
+);
+
```
TomJKing, line 159: Unnecessary line break

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepositorySpec.scala
```diff
@@ -195,4 +195,42 @@ class FileMetadataRepositorySpec extends AnyFlatSpec with TestDatabase with Scal
     fileTwoMetadata.get.head.propertyname should equal("FilePropertyOne")
     fileTwoMetadata.get.head.value should equal("Result of FileMetadata processing")
   }
+
+  "getClosureMetadataProperty" should "return the correct closure metadata property" in {
+    val db = DbConnection.db
+    val fileMetadataRepository = new FileMetadataRepository(db)
+    deleteTables()
```
TomJKing, line 202: Is there a reason for deleting the tables in this test and not the other tests that have been added?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -82,6 +96,52 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
...
+    (for {
+      properties <- fileMetadataRepository.getClosureMetadataProperty
+      values <- fileMetadataRepository.getClosureMetadataValues
+      dependencies <- fileMetadataRepository.getClosureMetadataDependencies
+    } yield (properties, values, dependencies)).map {
+      case (properties, valuesResult, dependenciesResult) =>
```
TomJKing, line 106: Should `properties` be renamed `propertiesResult` for consistency?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/MetadataFields.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.UUID
```
TomJKing, line 3: Unused imports that can be removed.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/MetadataFields.scala
```diff
@@ -0,0 +1,42 @@
...
+                          )
+
+  implicit val DataTypeType: EnumType[DataType] = deriveEnumType[DataType]()
+  implicit val PropertyTypeType: EnumType[PropertyType] = deriveEnumType[PropertyType]()
+  implicit val MetadataFieldsType: ObjectType[Unit, MetadataField] = deriveObjectType[Unit, MetadataField]()
+  implicit val MetadataValuesType: ObjectType[Unit, MetadataValues] = deriveObjectType[Unit, MetadataValues]()
+
+  val queryFields: List[Field[ConsignmentApiContext, Unit]] = fields[ConsignmentApiContext, Unit](
+    Field("getClosureMetadata", ListType(MetadataFieldsType),
+      arguments= Nil,
```
TomJKing, line 37: Missing space between `arguments` and `=`

src/main/scala/uk/gov/nationalarchives/tdr/api/auth/ValidationAuthoriser.scala
```diff
@@ -21,14 +21,6 @@ class ValidationAuthoriser(implicit executionContext: ExecutionContext)
                            ctx: Context[ConsignmentApiContext, _]
                           ): BeforeFieldResult[ConsignmentApiContext, Unit] = {
 
-    // All fields must have an authorisation tag defined. This means that if we forget to add authorisation, the
```
TomJKing, line 24: Probably a stupid idea, but, I'm wondering if we could actually pass in the consignment id, when calling the `getClosureMetadata`? Then check the user has access to the consignment (`ValidateUserHasAccessToConsignment`). I would think the consignment id is going to be available via the URL for the closure metadata form, eg `/consignmentid/fileid/closureMetadataForm`. At least then there would be some check. It would also stop a user from trying to update the custom metadata from another users upload?

@MancunianSam What do you think? I would quite like to keep this constraint.

TomJKing (commented):
Looks good. Just a few comments/suggestions. They might be incorrect, as not been very close to this bit of the work.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepositorySpec.scala
thanhz: Somewhere in the code the FilePropertyTable was being used which was causing this test to fail with an extremely vague error. @MancunianSam said at some point we are gonna remove the h2 database so most of the test will need to be redone anyway. unless I am mistaken.

TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):

PR #190, opened 2022-02-07
Move to github actions.

MancunianSam:
The flow is the same as it is for Jenkins. It will run the tests on push or pull request. There is a separate build job which runs on push to master which builds the code and the image and pushes them to ECR.

A lot of this will be common to a lot of the other test and deploy jobs. When I do another repository, I'll move the common code out into another repo but for now, I just want to get it working.

This is dependent on https://github.com/nationalarchives/tdr-e2e-tests/pull/166 which is itself dependent on a load of other stuff.


.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: github.ref == 'refs/heads/master'
+        run: |
+          export EXISTING_VERSION=$(curl ${{ github.api_url }}/repos/nationalarchives/tdr-consignment-api/tags | jq -r '.[0].name')
+          echo NEW_VERSION=$(echo "${EXISTING_VERSION:1}" | awk -F. '/[0-9]/{$NF+=1;OFS=".";print}' | xargs -I{} echo "v{}") >> $GITHUB_ENV
+      - name: Build new image version
+        if: github.ref == 'refs/heads/master'
+        env:
+          GITHUB_TOKEN: ${{ secrets.WORKFLOW_PAT }}
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
TomJKing, line 43: Is this a static user, that we've created, or something generated by GitHub actions?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Configure AWS credentials for ECS
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets[env.account_secret_name] }}:role/TDRGitHubECSUpdateRole${{ env.environment_title }}
+          aws-region: eu-west-2
+          role-session-name: APIECRLogin
+      - name: Deploy to ECS
+        run: aws ecs update-service --service consignmentapi_service_${{ github.event.inputs.environment }} --cluster consignmentapi_${{ github.event.inputs.environment }} --force-new-deployment
+      - name: Send service updated message
+        uses: slackapi/slack-github-action@v1.18.0
```
TomJKing, line 62: Can this version `v1.18.0` be stored as a param. It's also used in the test yaml?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,93 @@
+name: Deploy API
```
TomJKing, line 1: Not sure how this is displayed in GitHub actions, but maybe change the name to: `TDR Deploy API`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
+name: Run API tests
```
TomJKing, line 1: See comment about regarding the name.

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-latest
+    environment: intg
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: coursier/cache-action@v6
+      - name: Configure AWS credentials from management account
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.MANAGEMENT_ACCOUNT }}:role/GithubActionsRoleMgmt
```
TomJKing, line 19: Shouldn't our IAM role names be prefixed with `TDR`, ie `TDRGithubActionsRoleMgmt`

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+          role-session-name: APIDownloadDependencies
+      - name: Login to ECR
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ secrets.MANAGEMENT_ACCOUNT }}.dkr.ecr.eu-west-2.amazonaws.com
+      - name: Setup JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: temurin
+          java-version: 8
```
TomJKing, line 30: Are later versions of Java supported?

TomJKing (commented):
Looks good. Just some minor comments/questions. I think when the common stuff is pulled out into its own repo, will need a decent set of instructions in the README. maybe with some explanations of how this works. Some of the syntax used in the commands is not as refined as in the `Jenkinsfile`s

MancunianSam:
Marking this as draft as there's a load of other stuff that has to go in first

.github/workflows/deploy.yml
MancunianSam: It's just a name you see when you're viewing the actions in the browser so TDR Deploy API makes sense, I've changed it.

.github/workflows/deploy.yml
MancunianSam: Not really. Once I've got this working I'm going to have a proper look at creating some custom actions for these so we can reuse them and then it'll be in the same place then but for now I think it's ok to live with it.

.github/workflows/test.yml
MancunianSam: I've changed this one to TDR. There's a new build job as well which I've named with TDR as the prefix.

.github/workflows/test.yml
MancunianSam: This was the github actions bot but that doesn't have permissions to push to the repo and it shouldn't really anyway so I've switched this to use the existing user we use in Jenkins.

.github/workflows/test.yml
MancunianSam: Yeah, I've updated the terraform-environments and terraform-backend roles to be prefixed with TDR

.github/workflows/test.yml
MancunianSam: This isn't even necessary, I've taken it out.

TomJKing (approved):

PR #193, opened 2022-02-11
Tdr 1611 retrieve specific files metadata

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/AntivirusMetadataRepository.scala
```diff
@@ -18,10 +18,11 @@ class AntivirusMetadataRepository(db: Database)(implicit val executionContext: E
     db.run(allUpdates).map(_ => antivirusMetadataRow)
   }
 
-  def getAntivirusMetadata(consignmentId: UUID): Future[Seq[AvmetadataRow]] = {
+  def getAntivirusMetadata(consignmentId: UUID, selectedFileIds: Option[Set[UUID]]): Future[Seq[AvmetadataRow]] = {
```
MancunianSam, line 21: Is it worth setting the selectedFiles argument with a default value of `None` There are quite a few places where you're passing `None` in. Or are these eventually going to have a value passed to them in which case it's better to leave it without a default value.

MancunianSam:
Looks good. The selectedFiles aren't passed in from anywhere in the query yet, I guess that's coming later.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/AntivirusMetadataRepository.scala
TomJKing: Yeah, fair point. Updated to have a default of `None`

MancunianSam (approved):

PR #195, opened 2022-02-17
Add new confirmTransferStatus

thanhz:


techncl (changes requested):
Looks good just a few comments on the tests (I was unable to add the comments in the relevant place, so I'll make them here:

1. For this [test](https://github.com/nationalarchives/tdr-consignment-api/pull/195/files#diff-8af430ff536b04425b75b9adbd9c69938ea74433c1230ab729656c72206e0f29)

On line 29, could you add another consignment status row for the "Confirm transfer"

2. Also, in the `FinalConfirmationServiceSpec`, could you add a `addConfirmTransferStatus` test, similar to the `addTransferAgreementStatus` test in the `TransferAgreementServiceSpec`? Just to make sure that this method is doing the correct thing because now, the FinalConfirmationServiceSpec is only testing that the `addConsignmentMetadata` is passing the correct data to the repository method but we don't have a check/reassurance that the consignment status is being updated

PR #235, opened 2022-03-08
Add export version to consignment endpoint

techncl:


src/test/resources/scripts/init.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE IF NOT EXISTS Consignment (
   TransferInitiatedBy uuid,
   ExportDatetime timestamp with time zone,
   ExportLocation text,
+  ExportVersion text,
```
MancunianSam, line 34: This is the script for the old H2 database which we shouldn't need any more. I  meant to delete it. You can delete it as part of this PR or leave it for another time but you don't need to add the field in.

techncl:
> Would this be a breaking change? The export is expecting `updateExportLocation` and this is being removed and changed to `updateExportData` You can't deploy the export or the API first without breaking the export I don't think. Does that sound right?
> 
> If it is then we should add in a field for the existing mutation without the export version, either by putting in a static value or by changing the database to make the field nullable then changing it back once the export has switched over to use the new endpoint.
> 
> If I've missed something and we can deploy without breaking anything then ignore me.

Yes, you're right; this is a breaking change. Cool, I'll add a temporary default value

techncl:
I think I've figured out a solution to prevent any breaking. Let me know what you think.

MancunianSam (approved):
Looks good mate, once it's deployed everywhere we can remove the old endpoint.

PR #252, opened 2022-03-09
Include FFIDStatus response value

thanhz:


TomJKing (commented):
Looks good. I would update the getConsignment route spec tests to include the file status in the response as well for completeness.

I think its worth getting Sam to have a look at the PR, just to make sure he is happy with the overall approach to handling the file status.

MancunianSam:
Looks sensible to me, thanks.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -111,6 +111,7 @@ object FileMetadataService {
                   fileName: Option[String] = None,
                   parentId: Option[UUID] = None,
                   metadata: FileMetadataValues,
+                  filestatus: Option[String] = None,
```
MancunianSam, line 114: Just a small thing but can we have this as `fileStatus` instead of `filestatus` We're using camel case in other places for multiple words.

MancunianSam (approved):

PR #256, opened 2022-03-10
Add mergify yaml and scala steward config.

MancunianSam:
I've signed this repo up to get the scala steward updates. We also want
these to be auto merged because we don't want more dependency updates to
worry about.

I've added in the mergify config and got scalastyle to ignore the
consignment-api-db changes as there's an ancient version in maven.

I've also change the test trigger from pull_request to
pull_request_target. This means that the code for the github actions
always comes from master but this is fine because there's a push trigger
as well which will catch any errors if we do make changes to the action
files.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,20 @@
+queue_rules:
+  - name: default
+    conditions:
+      - check-success=test
+      - "check-success=security/snyk (nationalarchives)"
+pull_request_rules:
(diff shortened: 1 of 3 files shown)


TomJKing (approved):
Looks good. Assume you've done some due diligence on Mergify? The decision to use it should also go in an ADR I think, as its a sufficiently important change to justify that. Probably separate to the ADR for moving to GitHub actions, but the two can be linked together.

PR #265, opened 2022-03-15
Rename mergify check.

MancunianSam:
The name of the check is now test / test so needs to be updated to
allow mergify to work.

Needs https://github.com/nationalarchives/tdr-github-actions/pull/5
to be merged first.


TomJKing (approved):
Looks good.

PR #277, opened 2022-03-23
Fix mismatched file ids

MancunianSam:
Originally, we were creating files and then assigning an arbitrary file
ID to add the metadata. This worked because there was nothing in the
file table which identified the file it came from.

We now have a file name but we're still arbitrarily assigning IDs to the
FileMetadata table so sometimes, the file name and the original path
don't match up.

I've updated this to create the file rows and the fileMetadata rows
together with each loop through the values from the generateNodes
method.

I've added a test to check that the file name and original path match
and updated a couple of other tests where this has broken them.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
@@ -2,9 +2,10 @@ package uk.gov.nationalarchives.tdr.api.service
 
 import uk.gov.nationalarchives.Tables.{FileRow, FilemetadataRow}
 import uk.gov.nationalarchives.tdr.api.db.repository._
-import uk.gov.nationalarchives.tdr.api.graphql.fields.FileFields.{AddFileAndMetadataInput, ClientSideMetadataInput, FileMatches}
-import uk.gov.nationalarchives.tdr.api.model.file.NodeType.{FileTypeHelper, fileTypeIdentifier, directoryTypeIdentifier}
+import uk.gov.nationalarchives.tdr.api.graphql.fields.FileFields.{AddFileAndMetadataInput, FileMatches}
+import uk.gov.nationalarchives.tdr.api.model.file.NodeType.{FileTypeHelper, directoryTypeIdentifier, fileTypeIdentifier}
 import uk.gov.nationalarchives.tdr.api.service.FileMetadataService._
+import uk.gov.nationalarchives.tdr.api.service.FileService._
 import uk.gov.nationalarchives.tdr.api.utils.TimeUtils.LongUtils
 import uk.gov.nationalarchives.tdr.api.utils.TreeNodesUtils
 import uk.gov.nationalarchives.tdr.api.utils.TreeNodesUtils._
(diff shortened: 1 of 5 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -110,4 +104,24 @@ class FileService(
...
+    val fileRow: FileRow
+    val metadataRows: List[FilemetadataRow]
+  }
+
+  case class MatchedFileRows(fileRow: FileRow, metadataRows: List[FilemetadataRow], matchId: Long) extends Rows
+
+  case class DirectoryRows(fileRow: FileRow, metadataRows: List[FilemetadataRow]) extends Rows
+
+  case class FileOwnership(fileId: UUID, userId: UUID)
+}
```
TomJKing, line 127: Missing line break

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -42,37 +35,38 @@ class FileService(
...
 
-    val folderIdToPath: Map[UUID, String] = (allEmptyDirectoryNodes ++ allFileNodes.filter(_._2.treeNodeType == directoryTypeIdentifier))
-      .map(f => f._2.id -> f._1)
+    val row: (UUID, String, String) => FilemetadataRow = FilemetadataRow(uuidSource.uuid, _, _, now, userId, _)
 
-    val fileRows: List[FileRow] = ((allEmptyDirectoryNodes ++ allFileNodes) map {
-      case (_, treeNode) =>
+    val rowsWithMatchId: List[Rows] = ((allEmptyDirectoryNodes ++ allFileNodes) map {
+      case (path, treeNode) =>
         val parentId = treeNode.parentPath.map(v => allFileNodes.getOrElse(v, allEmptyDirectoryNodes(v)).id)
```
TomJKing, line 42: Maybe change the `v` param name to `path` just to make a bit clearer?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -2,8 +2,10 @@ package uk.gov.nationalarchives.tdr.api.routes
 
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import com.dimafeng.testcontainers.PostgreSQLContainer
+import io.circe.Printer
```
TomJKing, line 5: Is this import needed?

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -2,8 +2,10 @@ package uk.gov.nationalarchives.tdr.api.routes
...
 import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.auto._
+import io.circe.syntax.EncoderOps
```
TomJKing, line 8: Is this import needed?

TomJKing (commented):
Looks good. Just some very minor comments.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
MancunianSam: Makes sense, I've changed it.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
MancunianSam: No, I was printing the response json to see what was up with the tests. Same with the one below. I've removed them both.

PR #283, opened 2022-03-25
Tdr 1611 retrieve metadata file single call

TomJKing:


MancunianSam (approved):
Looks good mate. Some of those tests are pretty long with quite a bit of repetition. I had a go and moving stuff out into methods but it didn't make it much shorter and it was harder to read so I think it's fine.

PR #302, opened 2022-04-05
Add closure metadata support

techncl:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -31,35 +38,76 @@ class FileMetadataService(fileMetadataRepository: FileMetadataRepository,
...
+    } yield BulkFileMetadata(fileIdsAndMetadata._1.toSeq, fileIdsAndMetadata._2.toSeq)
+  }
+
+  def getFileMetadata(consignmentId: UUID): Future[Map[UUID, FileMetadataValues]] = fileMetadataRepository.getFileMetadata(consignmentId).map {
+    rows =>
+      rows.groupBy(_.fileid).map {
+        case (fileId, fileMetadata) =>
+          val propertyNameMap: Map[String, String] = fileMetadata.groupBy(_.propertyname)
+            .transform((_, value) => value.head.value)
+          fileId -> FileMetadataValues(
```
MancunianSam, line 98: Just a suggestion but we're doing something very similar here https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala#L104
That method in the FileService doesn't look like it needs to be in the implicit class either. Is it worth extracting the logic out somewhere to save repeating this to think.

src/test/resources/json/addbulkfilemetadata_data_fileids_missing.json
```diff
@@ -0,0 +1,14 @@
...
+      "message": "Variable '$addBulkFileMetadataInput' expected value of type 'AddBulkFileMetadataInput!' but got: {\"metadataProperties\":[{\"filePropertyName\":\"property1\",\"value\":\"value1\"},{\"filePropertyName\":\"property2\",\"value\":\"value2\"}]}. Reason: 'fileIds' Expected non-null value, found null. (line 1, column 30):\nmutation addBulkFileMetadata($addBulkFileMetadataInput: AddBulkFileMetadataInput!) {addBulkFileMetadata(addBulkFileMetadataInput:$addBulkFileMetadataInput) {fileIds,metadataProperties {filePropertyName,value}}}\n                             ^",
+      "locations": [
+        {
+          "column": 47,
+          "line": 1
+        }
+      ]
+    }
+  ]
+}
```
MancunianSam, line 14: Missing line break here.

src/test/resources/json/addbulkfilemetadata_data_metadataproperties_missing.json
```diff
@@ -0,0 +1,14 @@
...
+      "message": "Variable '$addBulkFileMetadataInput' expected value of type 'AddBulkFileMetadataInput!' but got: {\"fileIds\":[\"07a3a4bd-0281-4a6d-a4c1-8fa3239e1313\",\"d74650ff-21b1-402d-8c59-b114698a8341\",\"7076f399-b596-4161-a95d-e686c6435710\",\"d2e64eed-faff-45ac-9825-79548f681323\"]}. Reason: 'metadataProperties' Expected non-null value, found null. (line 1, column 30):\nmutation addBulkFileMetadata($addBulkFileMetadataInput: AddBulkFileMetadataInput!) {addBulkFileMetadata(addBulkFileMetadataInput:$addBulkFileMetadataInput) {fileIds,metadataProperties {filePropertyName,value}}}\n                             ^",
+      "locations": [
+        {
+          "column": 47,
+          "line": 1
+        }
+      ]
+    }
+  ]
+}
```
MancunianSam, line 14: Missing line break here too.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepositorySpec.scala
```diff
@@ -140,6 +140,26 @@ class FileMetadataRepositorySpec extends TestContainerUtils with ScalaFutures wi
...
+      val fileId = UUID.fromString("ba176f90-f0fd-42ef-bb28-81ba3ffb6f05")
+      utils.addFileProperty("FileProperty")
+      utils.createConsignment(consignmentId, userId)
+      utils.createFile(fileId, consignmentId)
+      val numberOfFileMetadataRows = 100
+      val input = (1 to numberOfFileMetadataRows).map(_ =>
+        FilemetadataRow(UUID.randomUUID(), fileId, "value", Timestamp.from(Instant.now()), UUID.randomUUID(), "FileProperty")
+      )
+
+      val result = fileMetadataRepository.addFileMetadata(input).futureValue
```
MancunianSam, line 158: This result variable looks like it's unused.

MancunianSam (commented):
Looks good, just a couple of things.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
techncl: Yes, I agree although, even though it looks like I added this method, I didn't actually do it; I think it looks that way due to an merge conflict issue, it seems as though @TomJKing created both. @TomJKing, is there a reason why you added an implicit class to the [this](https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala#L104)?

TomJKing: I think I just missed this when I implemented getting the `FileRow` and `FilemetadataRow` with this method in the `FileService` at line 82 (another facepalm moment...)

The implicit class was there to make things a bit easier. 

We should probably be using the `FileService` method and not have two methods doing essentially the same thing? What do you think, the only difference in the return type is that whole `FileRow` is returned instead of just the `FileId`. Whatever is consuming this response should be able to handle this, and do what it needs. Happy to discuss if I'm talking complete nonesense.

techncl: 👍 . Cool, I've pulled the duplicate code out into a method.

MancunianSam (approved):

PR #303, opened 2022-04-05
Fix the file metadata matching.

MancunianSam:
This line
`val pathWithoutSlash = if (m.originalPath.startsWith("/")) m.originalPath.tail else path`
seems to have been put in place in case we get a forward slash sent to the API.

The problem was that the originalPath sent from the front end has no leading slash so it was returning the calculated input `path` which we were then comparing with the calculated input `path` which always returned true.

We then called head on the list and got the same object each time.

This returned the same match id to the front end for both files so the front end uploaded the same file twice.

We have changed the tests so they match the real inputs. They have no leading slash.

We may not even need this check for the forward slash but we don't want to remove it now in case it handles an edge case somewhere.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       repo-name: tdr-consignment-api
       test-command: |
         docker build -f Dockerfile-tests -t tests .
-        sbt test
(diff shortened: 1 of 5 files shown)


TomJKing (approved):
Looks good. Thanks

PR #308, opened 2022-04-14
Add 'export' status to CurrentStatus and add Export Status row when t…

techncl:
…ransfer starts

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
@@ -46,11 +46,11 @@ class ConsignmentService(
   }
 
   def updateTransferInitiated(consignmentId: UUID, userId: UUID): Future[Int] = {
-    val updateTransferInitiatedStatus = consignmentRepository.updateTransferInitiated(consignmentId, userId, Timestamp.from(timeSource.now))
-    //Later, "Completed" might be changed to "InProgress"
-    val consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, "Export", "Completed", Timestamp.from(timeSource.now))
-    consignmentStatusRepository.addConsignmentStatus(consignmentStatusRow)
-    updateTransferInitiatedStatus
+    for {
+      updateTransferInitiatedStatus <- consignmentRepository.updateTransferInitiated(consignmentId, userId, Timestamp.from(timeSource.now))
+      consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, "Export", "Completed", Timestamp.from(timeSource.now))
+      _ <- consignmentStatusRepository.addConsignmentStatus(consignmentStatusRow)
+    } yield updateTransferInitiatedStatus
   }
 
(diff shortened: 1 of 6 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -46,7 +46,11 @@ class ConsignmentService(
   }
 
   def updateTransferInitiated(consignmentId: UUID, userId: UUID): Future[Int] = {
-    consignmentRepository.updateTransferInitiated(consignmentId, userId, Timestamp.from(timeSource.now))
+    val updateTransferInitiatedStatus = consignmentRepository.updateTransferInitiated(consignmentId, userId, Timestamp.from(timeSource.now))
+    //Later, "Completed" might be changed to "InProgress"
+    val consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, "Export", "Completed", Timestamp.from(timeSource.now))
+    consignmentStatusRepository.addConsignmentStatus(consignmentStatusRow)
+    updateTransferInitiatedStatus
   }
```
MancunianSam, line 54: Can you flatMap these two together. As it is, this will work because `Future` is eager but we could end up in a state where `updateTransferInitiated` fails but `addConsignmentStatus` succeeds. If you flatMap the two together, a bit like this:

```scala
for {
      updateTransferInitiatedStatus <- consignmentRepository.updateTransferInitiated(consignmentId, userId, Timestamp.from(timeSource.now))
      consignmentStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, "Export", "Completed", Timestamp.from(timeSource.now))
      _ <- consignmentStatusRepository.addConsignmentStatus(consignmentStatusRow)
    } yield updateTransferInitiatedStatus
```
That should sort it.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -12,8 +12,13 @@ import org.scalatest.flatspec.AnyFlatSpec
...
 import uk.gov.nationalarchives.tdr.api.db.repository._
-import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields.{AddConsignmentInput, FileChecks, StartUploadInput,
-  UpdateConsignmentSeriesIdInput, UpdateExportDataInput}
+import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields.{
+  AddConsignmentInput,
+  FileChecks,
+  StartUploadInput,
+  UpdateConsignmentSeriesIdInput,
+  UpdateExportDataInput
+}
```
MancunianSam, line 21: I guess this is scalastyle complaining about the length of the line. We don't have many of these multi line imports so would it be better to just do `import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields._`

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
techncl: That's a very good point.

MancunianSam (approved):

PR #312, opened 2022-04-20
Make 'setUploadConsignmentStatusValueToComplete' more generic

techncl:
...by making it an 'updateConsignmentStatus' method.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentStatusFields.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentStatusFields.scala
@@ -16,7 +16,11 @@ object ConsignmentStatusFields {
   val mutationFields: List[Field[ConsignmentApiContext, Unit]] = fields[ConsignmentApiContext, Unit](
     Field("markUploadAsCompleted", OptionType(IntType),
       arguments = ConsignmentIdArg :: Nil,
-      resolve = ctx => ctx.ctx.consignmentStatusService.setUploadConsignmentStatusValueToComplete(ctx.arg(ConsignmentIdArg)),
+      resolve = ctx => ctx.ctx.consignmentStatusService.updateConsignmentStatus(
+        ctx.arg(ConsignmentIdArg),
+        "Upload",
+        "Completed"
+      ),
       tags = List(ValidateUserHasAccessToConsignment(ConsignmentIdArg))
     )
   )
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentStatusFields.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentStatusFields.scala
@@ -1,18 +1,30 @@
 package uk.gov.nationalarchives.tdr.api.graphql.fields
 
 import io.circe.generic.auto._
+import sangria.macros.derive.deriveInputObjectType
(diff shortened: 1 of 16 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentStatusFields.scala
```diff
@@ -16,7 +16,11 @@ object ConsignmentStatusFields {
   val mutationFields: List[Field[ConsignmentApiContext, Unit]] = fields[ConsignmentApiContext, Unit](
     Field("markUploadAsCompleted", OptionType(IntType),
       arguments = ConsignmentIdArg :: Nil,
-      resolve = ctx => ctx.ctx.consignmentStatusService.setUploadConsignmentStatusValueToComplete(ctx.arg(ConsignmentIdArg)),
+      resolve = ctx => ctx.ctx.consignmentStatusService.updateConsignmentStatus(
+        ctx.arg(ConsignmentIdArg),
```
TomJKing, line 20: Hardcoding these values should be a temporary measure. We should:
1. create new generic mutation: `updateConsignmentStatus` which takes the relevant arguments
2. commit these changes
3. move any clients using the old mutation: `markUploadAsCompleted`
4. remove the old mutation

TomJKing (commented):
Looks good. I've added a comment about creating a new generic mutation. Thanks

TomJKing (commented):
Looks good. One thing that has occurred to me, is that we should probably be checking the `statusType` and `statusValue` inputs are valid.

Potentially create a companion object in the `ConsignmentStatusService.scala` and do something like this:

```
object ConsignmentStatusService {
  val exportStatusType = "Export"
  val exportStatusValues = Set("Upload", "Completed")

  val consignmentStatuses: Map[String, Set[String]] = Map(exportStatusType -> exportStatusValues)

  def validateConsignmentStatus(statusType: String, statusValue: String): Boolean = {
    if(consignmentStatuses.contains(statusType)) {
      consignmentStatuses.get(statusType).fold(
        throw InputDataException(s"Invalid consignment status value '$statusValue' for status type '$statusType'"))(_.contains(statusValue))
    } else {
      throw InputDataException(s"Invalid consignment status type '$statusType'")
    }
  }
}
```
Sorry, only thought of this now.

.bsp/sbt.json
```diff
@@ -0,0 +1 @@
+{"name":"sbt","version":"1.6.2","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Library/Java/JavaVirtualMachines/jdk-15.0.2.jdk/Contents/Home/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/nlutale/Library/Application Support/JetBrains/IntelliJIdea2021.3/plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp","--sbt-launch-jar=/Users/nlutale/Library/Application%20Support/JetBrains/IntelliJIdea2021.3/plugins/Scala/launcher/sbt-launch.jar"]}
```
TomJKing, line 1: This file shouldn't be merged. Should probably be added to the `.gitignore` file

TomJKing (commented):
Looks good. Just one minor comment.

.bsp/sbt.json
techncl: 🤦‍♂️ , I did a `git add .`

TomJKing (approved):
Looks good. Thanks

PR #337, opened 2022-05-18
Add FileService method to return basic paginated files response

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -91,8 +94,26 @@ class FileService(fileRepository: FileRepository,
...
     } yield fileAndMetadataList.toFiles(avList, ffidMetadataList, ffidStatus).toList
   }
-}
 
+  def getPaginatedFiles(consignmentId: UUID,
+                        paginationInput: PaginationInput,
+                        fileFilters: Option[FileFilters] = None): Future[List[File]] = {
+    val filters = fileFilters.getOrElse(FileFilters())
+    val inputCursor = paginationInput.currentCursor
+    val currentCursor = if (inputCursor.isDefined) Some(UUID.fromString(inputCursor.get)) else None
```
MancunianSam, line 103: This is completely my opinion so feel free to ignore but you can simplify this a bit 

`val currentCursor = inputCursor.map(UUID.fromString)`

MancunianSam (approved):

PR #350, opened 2022-05-27
Tdr 1611 paginated files retrieve metadata

TomJKing:


MancunianSam (dismissed):
Looks good, thanks.

MancunianSam (approved):
You had a merge conflict on the auth url which I just set to what was on master.

PR #357, opened 2022-06-08
One container per test

MancunianSam:
This runs one database docker container per test which is quite slow so now we're running them in parallel


TomJKing (approved):
Looks good. Thanks

PR #377, opened 2022-07-22
Update consignment query

vimleshtna:
Add file filters in consignment's 'files' field to query the files only for given file Ids

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
@@ -2,7 +2,6 @@ package uk.gov.nationalarchives.tdr.api.graphql.fields
 
 import java.time.ZonedDateTime
 import java.util.UUID
-
 import akka.http.scaladsl.server.RequestContext
 import io.circe.generic.auto._
 import sangria.macros.derive._
@@ -11,6 +10,7 @@ import sangria.relay._
 import sangria.schema.{Argument, BooleanType, Field, InputObjectType, IntType, ListType, ObjectType, OptionInputType, OptionType, StringType, fields}
 import uk.gov.nationalarchives.tdr.api.auth._
 import uk.gov.nationalarchives.tdr.api.consignmentstatevalidation.ValidateNoPreviousUploadForConsignment
+import uk.gov.nationalarchives.tdr.api.db.repository.FileFilters
 import uk.gov.nationalarchives.tdr.api.graphql._
 import uk.gov.nationalarchives.tdr.api.graphql.fields.FieldTypes._
 import uk.gov.nationalarchives.tdr.api.graphql.validation.UserOwnsConsignment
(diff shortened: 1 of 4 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -134,7 +136,11 @@ object ConsignmentFields {
       Field(
         "files",
         ListType(FileType),
-        resolve = context => DeferFiles(context.value.consignmentid)
+        arguments = FileFiltersInputArg :: Nil,
```
TomJKing, line 139: The `fileFiltersInputArg` will need to be passed into the `getConsignment` query (line 188) as an optional argument I think, otherwise it won't be accessible by the deferred resolve for getting the files.
...
```
      arguments = ConsignmentIdArg :: PaginationInputArg :: FileFiltersInputArg :: Nil,
      ...
    )
...
```

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
```diff
@@ -229,6 +229,30 @@ class FileRepositorySpec extends TestContainerUtils with ScalaFutures with Match
       files.size shouldBe 1
   }
 
+  "getFiles" should "return files and file metadata where 'selectedFileIds' filter applied" in withContainers {
```
TomJKing, line 232: Would reword the test slightly to: `... return files and file metadata for selected files where 'selectedFileIds' ...`

I would also add the directory to the selected files list to show it will bring back both types where the type filter is not applied.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
```diff
@@ -229,6 +229,30 @@ class FileRepositorySpec extends TestContainerUtils with ScalaFutures with Match
...
+      val utils = TestUtils(db)
+      val fileRepository = new FileRepository(db)
+      val consignmentId = UUID.fromString("c6f78fef-704a-46a8-82c0-afa465199e66")
+      setUpFilesAndDirectories(consignmentId, utils)
+
+      val files = fileRepository.getFiles(consignmentId, FileFilters(selectedFileIds = Some(List(fileOneId)))).futureValue
+      files.forall(_._1.fileid == fileOneId) shouldBe true
+  }
+
+  "getFiles" should "return files and file metadata where and 'file' type and 'selectedFileIds' filter applied" in withContainers {
```
TomJKing, line 244: Would reword the test slightly to: `... return files and file metadata for selected files where 'file' type and 'selectedFileIds' filters are applied`

TomJKing (commented):
Looks good. Some minor comments. Also the `ConsignmentRouteSpec` tests should be updated with some tests with using the "file filters" you've added.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
vimleshtna: Actually, I tested and it works fine. I think we don't need to pass with `getConsignment` 2field as we are using the `fileFiltersInputArg` in the `files` field only

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
vimleshtna: Sure, I'll add it too. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
TomJKing: As per our conversation, yes correct. Please ignore this comment.

TomJKing (approved):
Looks good. Thanks.

PR #379, opened 2022-07-26
Tdr 2119 File Navigation API

thanhz:


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
```diff
@@ -58,6 +58,17 @@ class FileRepository(db: Database)(implicit val executionContext: ExecutionConte
     db.run(query.result)
   }
 
+  def countFilesOrFoldersInConsignment(consignmentId: UUID,
```
TomJKing, line 61: I think it would be safe to change the existing `countFilesInConsignment` method, as below and use this for the pagination counts:

```
def countFilesInConsignment(consignmentId: UUID,
                              parentId: Option[UUID] = None,
                              fileTypeIdentifier: Option[String] = Some(NodeType.fileTypeIdentifier)): Future[Int] = {
    val query = File.filter(_.consignmentid === consignmentId)
      .filterOpt(fileTypeIdentifier)(_.filetype === _)
      .filterOpt(parentId)(_.parentid === _)
      .length
    db.run(query.result)
  }
```
We want to try and avoid creating more methods than are needed, as we already have a lot, doing very similar things. You can re-use the existing unit tests you'd written for the new count method, just including a new one where the `fileTypeIdentifier` is passed in as `None`.

Also I think for the pagination count I think you'll want to get back all the direct children nodes for a given parent id, to provide the correct count to the UI for the display, which would mean using a `fileTypeIdentifier` of `None`

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
```diff
@@ -81,13 +92,17 @@ class FileRepository(db: Database)(implicit val executionContext: ExecutionConte
 
   def getPaginatedFiles(consignmentId: UUID,
                         limit: Int,
-                        after: Option[UUID],
+                        offset: Int,
+                        after: Option[String],
                         fileFilters: FileFilters): Future[Seq[FileRow]] = {
+    val selectedFileIds = fileFilters.selectedFileIds.getOrElse(List())
```
TomJKing, line 98: I would add a new `parentId` field to the `FileFilters` case class, and use this instead of the `selectedFileIds`. This will make it more explicit what is being filtered on:

`case class FileFilters(fileTypeIdentifier: Option[String] = None, selectedFileIds: Option[List[UUID]] = None, parentId: Option[UUID] = None)`

This will also allow you to pass it directly into the `.filterOpt`:
```
...
.filterOpt(fileFilters.parentId)(_.parentid === _)
...

```

TomJKing (commented):
Looks good. Mostly minor comments, and some suggested refactoring to simplify things as much as possible.

src/main/resources/application.base.conf
```diff
@@ -22,7 +22,7 @@ consignmentapi {
 
 pagination {
     consignmentsMaxLimit = 250
-    filesMaxLimit = 250
+    filesMaxLimit = 10000
```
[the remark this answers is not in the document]
thanhz: ah yes I forgot about that

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepository.scala
thanhz: This makes sense. The `getPaginatedFiles` method currently calls the `countFilesInConsignment` method with a parameter which could be `None`. So the default value won't be used.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileRepositorySpec.scala
```diff
@@ -87,6 +87,86 @@ class FileRepositorySpec extends TestContainerUtils with ScalaFutures with Match
...
+      utils.createFile(UUID.fromString("2dfa0495-72a3-4e88-9c0e-b105d7802a4e"), consignmentId)
+      utils.createFile(UUID.fromString("1ad53749-aba4-4369-8fd6-2311111427cc"), consignmentId)
+
+      val consignmentFiles = fileRepository.countFilesInConsignment(consignmentId,
+        fileTypeIdentifier = Some(NodeType.fileTypeIdentifier)).futureValue
+
+      consignmentFiles shouldBe 3
+  }
+
+  "countFilesInConsignment" should "return the total number of files in a consignment given a parentId filter" in withContainers {
```
TomJKing, line 151: This test should be rename: `... "return the total number of files and folders with the same parent given a parentId filter"`

TomJKing (commented):
Looks good. Just one very minor additional comment. Thanks

TomJKing (approved):
Looks good. Thanks.

PR #380, opened 2022-07-27
Parallelise some repo calls

techncl:
Some of the for comprehension repo method calls were sequential when they didn't need to be so this PR makes the repo calls run in parallel. By using a crude non-scientific method (`(new java.util.Date()).getTime`) in the RouteSpec tests, an up to 40% decrease in the time it takes to retrieve the data was witnessed.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
```diff
@@ -16,10 +16,13 @@ class FileStatusService(fileStatusRepository: FileStatusRepository)(implicit val
   }
 
   def allChecksSucceeded(consignmentId: UUID): Future[Boolean] = {
+    val getChecksumMatchStatus = fileStatusRepository.getFileStatus(consignmentId, ChecksumMatch)
+    val getAvStatus = fileStatusRepository.getFileStatus(consignmentId, Antivirus)
+    val getFfidStatus = fileStatusRepository.getFileStatus(consignmentId, FFID)
```
thanhz, line 21: Can these be put inside the for loop similarly to how you've done it for the others, for consistency?

thanhz (commented):
Looks good! parallelism ftw :+1:

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -92,10 +92,13 @@ class FileService(fileRepository: FileRepository,
     val filters = fileFilters.getOrElse(FileFilters())
     for {
       fileAndMetadataList <- fileRepository.getFiles(consignmentId, filters)
-      ffidMetadataList <- ffidMetadataService.getFFIDMetadata(consignmentId)
-      avList <- avMetadataService.getAntivirusMetadata(consignmentId)
-      ffidStatus <- fileStatusService.getFileStatus(consignmentId)
-    } yield fileAndMetadataList.toFiles(avList, ffidMetadataList, ffidStatus).toList
+      getFfidMetadataList: Future[List[FFIDMetadata]] = ffidMetadataService.getFFIDMetadata(consignmentId)
+      getAvList: Future[List[AntivirusMetadata]] = avMetadataService.getAntivirusMetadata(consignmentId)
+      getFfidStatus: Future[Map[UUID, String]] = fileStatusService.getFileStatus(consignmentId)
```
thanhz, line 97: Can these be put outside of the for loop? similarly to how you've done it for the other files. Just for consistency.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -107,14 +110,17 @@ class FileService(fileRepository: FileRepository,
...
       response: Seq[FileRow] <- fileRepository.getPaginatedFiles(consignmentId, maxFiles, currentCursor, filters)
       fileIds = Some(response.map(_.fileid).toSet)
-      fileMetadata <- fileMetadataService.getFileMetadata(consignmentId, fileIds)
-      ffidMetadataList <- ffidMetadataService.getFFIDMetadata(consignmentId, fileIds)
-      avList <- avMetadataService.getAntivirusMetadata(consignmentId, fileIds)
-      ffidStatus <- fileStatusService.getFileStatus(consignmentId, fileIds)
+      getFileMetadata = fileMetadataService.getFileMetadata(consignmentId, fileIds)
+      getFfidMetadataList = ffidMetadataService.getFFIDMetadata(consignmentId, fileIds)
+      getAvList = avMetadataService.getAntivirusMetadata(consignmentId, fileIds)
+      getFfidStatus = fileStatusService.getFileStatus(consignmentId, fileIds)
```
thanhz, line 119: Similar to the other comment.

thanhz (commented):
Just a couple of minor points

TomJKing (commented):
We just need to be aware of the impact on the DB performance with making some of the calls parallel. Its a bottleneck at the moment, so this may exacerbate  this. We can try it, and see what happens.

It'll make the argument stronger for upgrading the DB to something more powerful.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
techncl: Hi Thanh, I wanted to but the issue with this is that `avList`, `ffidMetadataList`, `ffidStatus` are passed into the `tofiles` method of `fileAndMetadataList`; I could be wrong but I think that we'd need to ensure that `fileAndMetadataList` is completed beforehand.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
techncl: Same comment as the one I made above; I think we need to ensure that `fileAndMetadataList` is completed first.

thanhz:
In addition the Tom's comment on the db being a bottleneck, maybe we should wait for the database index to go in and the changes we were going to implement for the database/graphql first before we merge this.

TomJKing:
@techncl Shall we park this for now, and close this PR, as not sure when we will get a chance to look at this properly.

techncl:
Sure.

PR #390, opened 2022-08-08
Add ordinal to metadata query

MancunianSam:
This returns the ordinal value from the database where it's set. If it's not set, it returns `Int.MaxValue` This should make sure that any unordered fields show up at the end and simplifies the sorting on the client.


vimleshtna (approved):
Looks good, thanks.

PR #391, opened 2022-08-08
Add a total items field.

MancunianSam:
In this query, it is populated with total files which is used on the
front end to display the total number of files for that folder.

I kept the name slightly more generic because this case class can be
used for other non-file things.

Instead of writing whole new tests I've added a couple of assertions
about the new totalItems field.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
@@ -119,7 +119,7 @@ class FileService(fileRepository: FileRepository,
 
     for {
       response: Seq[FileRow] <- fileRepository.getPaginatedFiles(consignmentId, maxFiles, offset, currentCursor, filters)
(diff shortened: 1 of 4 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -138,6 +138,7 @@ class FileService(fileRepository: FileRepository,
           hasPreviousPage = currentCursor.isDefined
         ),
         edges,
+        numberOfFilesInFolder,
```
thanhz, line 141: I know I named this. But maybe it should be renamed to `totalItems` for consistency?

thanhz (commented):
Looks good thanks! Just one small thing

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
MancunianSam: Ah good point yeah, I've renamed it.

thanhz (approved):

PR #401, opened 2022-08-25
TDR-2110 - File Level add new status property ClientChecksumStatus

vimleshtna:


src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -401,11 +406,22 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
         case _ => count should equal(2)
       }
     })
+    verify(fileStatusRepositoryMock, times(2)).addFileStatuses(any[List[FilestatusRow]]())
```
MancunianSam, line 409: This is a duplicate of the verify call on line 379. You can get rid of one of them.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -461,6 +479,20 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
...
```
MancunianSam, line 482: You've got two calls to `verify` in this test as well, you can delete one.

MancunianSam (commented):
Looks good thanks. There's just a couple of duplicate verify calls in your tests.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
vimleshtna: Removed. Thanks for highlighting it, probably it's a copy-paste error

MancunianSam (approved):

PR #403, opened 2022-08-26
Return an error for non judgment files

MancunianSam:
The allChecksSucceeded method was only returning an error for the FFID
status if it was contained in DisallowedPuids but that table doesn't
have NonJudgmentFormat in the table so if the status was
NonJudgmentFormat, it was returning allChecksSucceeded=true.

This checks for the statuses in that table and the NonJudgmentFormat.
I've added in a unit test to check for it.


thanhz (commented):
Looks good, Just wondering if it would be better to just add a `NonJudgmentFormat` to the `disallowedPuids `table? similar to how `ZeroByteFile` is added

MancunianSam:
@thanhz That's a good point, thanks. I didn't know we were doing that with zero byte files. and it would be better if this was data driven. I'll close this then as I don't think I need it and put in a PR to add a row to the `DisallowedPuids` table.

MancunianSam:
https://github.com/nationalarchives/tdr-consignment-api-data/pull/186 is the PR for the database update.

PR #409, opened 2022-09-06
TDR-2288 - order files/folders by natura order for navigation page

vimleshtna:
Scala doesn't have a built-in function to sort by natural order. We found one working solution for natural sorting and used it for files and folders


src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.util.matching.Regex
+import java.text.Normalizer
+
+object NaturalSorting {
+  implicit object ArrayOrdering extends Ordering[Array[String]] { // 4
+    val INT: Regex = "([0-9]+)".r
+    def compare(a: Array[String], b: Array[String]) : Int = {
+      val l = Math.min(a.length, b.length)
+      (0 until l).segmentLength(i => a(i) equals b(i)) match {
+        case i if i == l => Math.signum(b.length - a.length).toInt
```
MancunianSam, line 12: This line means that if you have fileName fileName2 fileName10 fileName3, it will sort them as fileName2 fileName3 fileName10 fileName. 
Most of the natural ordering sites online as well as using ls -v will put fileName first which makes sense I think. You can get it to sort like that just by swapping this around so it's `Math.signum(a.length - b.length).toInt`

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -805,6 +805,47 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
...
+
+    val fileRowParams = List(
+      (UUID.randomUUID(), consignmentId, "fileName2", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName22", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName21", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName31", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName32", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName1", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName5", parentId),
+      (UUID.randomUUID(), consignmentId, "fileName3", parentId)
```
MancunianSam, line 821: Could you put a "fileName" in here without a digit so we can check it comes first in the list.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def natural(str: String): Array[String] = {
+    val replacements = Map('\u00df' -> "ss", '\u017f' -> "s", '\u0292' -> "s").withDefault(s => s.toString) // 8
+    Normalizer.normalize(Normalizer.normalize(
+      str.trim.toLowerCase, // 1.1, 1.2, 3
+      Normalizer.Form.NFKC), // 7
+      Normalizer.Form.NFD).replaceAll("\\p{InCombiningDiacriticalMarks}", "") // 6
+      .replaceAll("^(the|a|an) ", "") // 5
+      .flatMap(replacements.apply) // 8
+      .split(s"\\s+|(?=[0-9])(?<=[^0-9])|(?=[^0-9])(?<=[0-9])") // 1.3, 2 and 4
```
MancunianSam, line 29: I don't think you need these comments here.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
```diff
@@ -0,0 +1,31 @@
+package uk.gov.nationalarchives.tdr.api.utils
+
+import scala.util.matching.Regex
+import java.text.Normalizer
+
+object NaturalSorting {
+  implicit object ArrayOrdering extends Ordering[Array[String]] { // 4
```
MancunianSam, line 7: I think you found this [on this site](https://rosettacode.org/wiki/Natural_sorting#Scala) Could you put a comment in with a link back to it. It's quite a difficult thing to follow and I think it would help to have the context from the site.

MancunianSam (commented):
Looks good thanks. Just the thing about sorting without numbers and  some unused comments.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
vimleshtna: Yeah, right that's how it displays on the file explorer as well. I have swapped the variables as you suggested and it works well. Thanks.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
vimleshtna: Added it and it is working as expected after the change on the `NatuaralSorting` object as suggested in the above comment. Thanks.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
vimleshtna: Removed it, thanks.

src/main/scala/uk/gov/nationalarchives/tdr/api/utils/NaturalSorting.scala
vimleshtna: Yeah, that make sense. Add in the doc comment.

MancunianSam (approved):

PR #414, opened 2022-09-08
Add a line to the merge strategy.

MancunianSam:
SLF4J needs to check the META-INF services directory to find the logging
implementation but it can't because the assembly plugin deletes it.

This keeps the services directory and now we have logging again.


thanhz (approved):
Thanks! CloudWatch logs showing up again :+1:

PR #423, opened 2022-09-21
TDR-2317 - Get consignment by user id

vimleshtna:
If the user has reporting access then return all the consignments else return the current user’s consignments only

Please have a look and let me know your thoughts. I'm not sure if this is good enough to validate and return the consignments.


TomJKing:
I think we can still use `ValidateHasReportingAccess` check, and:

1. pass the user id into the query;
2. use the user id to then check it matches the user id on the token. There is access to the arguments passed into a query on the context object in the validate functions. See for example: https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala#L39

See the psuedo code below. That way, we're still using a validation check, and a user should only have access to their own consignments through the query. 

```
...
consignments(... userId: UUID) { 
  ...
  tags = List(ValidateHasReportingAccess)
}

...

//probably rename this to something like: ValidateHasConsignmentsAccess
object ValidateHasReportingAccess extends SyncAuthorisationTag {
  override def validateSync(ctx: Context[ConsignmentApiContext, _]): BeforeFieldResult[ConsignmentApiContext, Unit] = {
    val userId = ctx.arg("userId")
    val token = ctx.ctx.accessToken
    val reportingAccess = token.reportingRoles.contains(reportingRole)
    if (reportingAccess || userId == token.userId) {
      continue
    } else {
      val tokenUserId = token.userId
      throw AuthorisationException(s"User $tokenUserId does not have permission to run reporting")
    }
  }
}

...
```

vimleshtna:
> ctx.arg

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -91,11 +91,11 @@ class ConsignmentService(
       row => convertRowToConsignment(row)))
   }
 
-  def getConsignments(limit: Int, currentCursor: Option[String]): Future[PaginatedConsignments] = {
+  def getConsignments(limit: Int, currentCursor: Option[String], consignmentFilter: Option[ConsignmentFilter] = None): Future[PaginatedConsignments] = {
     val maxConsignments: Int = min(limit, maxLimit)
 
     for {
-      response <- consignmentRepository.getConsignments(maxConsignments, currentCursor)
+      response <- consignmentRepository.getConsignments(maxConsignments, currentCursor, consignmentFilter.flatMap(_.userId))
```
TomJKing, line 98: See about about passing in the "consignment filters" argument to the repository method.

src/test/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepositorySpec.scala
```diff
@@ -232,6 +233,46 @@ class ConsignmentRepositorySpec extends TestContainerUtils with ScalaFutures wit
       response should have size 0
   }
 
+  "getConsignments" should "return all the consignments which assigned to the given user id only" in withContainers {
```
TomJKing, line 236: I would reword the test slightly to: `... consignments which belong to the given user id ...`

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -643,6 +643,86 @@ class ConsignmentRouteSpec extends TestContainerUtils with Matchers with TestReq
       response.data should equal(expectedResponse.data)
   }
 
+  "consignments" should "allow a user without reporting access to return only his consignments" in withContainers {
```
TomJKing, line 646: I would reword the test title slightly to: `... to return on their consignments`

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -643,6 +643,86 @@ class ConsignmentRouteSpec extends TestContainerUtils with Matchers with TestReq
...
+
+      val userAccessToken = validUserTokenNoBody
+
+      val expectedResponse: GraphqlConsignmentsQueryData = expectedConsignmentsQueryResponse("data_userId")
+      val response: GraphqlConsignmentsQueryData = runConsignmentsTestQuery("query_with_userId", userAccessToken)
+
+      response.data.get.consignments should equal(expectedResponse.data.get.consignments)
+  }
+
+  "consignments" should "not allow a user without reporting access, can access the consignments without passing userId" in withContainers {
```
TomJKing, line 675: I would reword the test title slightly to: `... reporting access to access consignments without ...`

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -643,6 +643,86 @@ class ConsignmentRouteSpec extends TestContainerUtils with Matchers with TestReq
...
+
+      val userAccessToken = validUserTokenNoBody
+
+      val response: GraphqlConsignmentsQueryData = runConsignmentsTestQuery("query_no_userId", userAccessToken)
+
+      response.errors should have size 1
+      response.errors.head.extensions.get.code should equal("NOT_AUTHORISED")
+  }
+
+  "consignments" should "not allow a user without reporting access, can access others consignments" in withContainers {
```
TomJKing, line 697: I would reword the test title slightly to: `... without reporting access to access other users' consignments`

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -451,6 +453,42 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
     exportLocations should contain (exportLocation2)
   }
 
+  "getConsignments" should "return all the consignments when consignment filter is passed" in {
```
TomJKing, line 456: Should all consignment be returned when there is no consignment filter? When there is a consignment filter the results would be filtered based on the arguments to the filter?

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -266,4 +273,18 @@ object ConsignmentFields {
...
     )
   )
+
+  private def getConsignmentsArgs(ctx: Context[ConsignmentApiContext, Unit]): (Int, Option[String], Option[ConsignmentFilter]) = {
+
+    val userId: Option[UUID] = ctx.args.argOpt("userId")
+    val limit: Int = ctx.args.arg("limit")
+    val currentCursor = ctx.args.argOpt("currentCursor")
+    val reportingAccess = ctx.ctx.accessToken.reportingRoles.contains(reportingRole)
+    val consignmentFilter = if (reportingAccess) {
```
TomJKing, line 283: Even if a user has "reporting access", they still should also have the ability to filter by a user id if they need to. Doing this will prevent that.

TomJKing (commented):
Looks good. Just some questions/suggestions. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -211,11 +219,10 @@ object ConsignmentFields {
       tags = List(ValidateUserHasAccessToConsignment(ConsignmentIdArg))
     ),
     Field("consignments", consignmentConnections,
-      arguments = List(LimitArg, CurrentCursorArg),
+      arguments = List(LimitArg, CurrentCursorArg, UserIdArg),
```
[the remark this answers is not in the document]
vimleshtna: Good point, I have used the `ConsignmentFilters` as an argument

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -266,4 +273,18 @@ object ConsignmentFields {
       tags = List(ValidateUserHasAccessToConsignment(UpdateConsignmentSeriesIdArg), ValidateUpdateConsignmentSeriesId)
     )
   )
+
+  private def getConsignmentsArgs(ctx: Context[ConsignmentApiContext, Unit]): (Int, Option[String], Option[ConsignmentFilter]) = {
```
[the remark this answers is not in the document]
vimleshtna: I have removed it and passed the consignment filter to the service directly.

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -55,8 +55,9 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
     db.run(query.result)
   }
 
-  def getConsignments(limit: Int, after: Option[String]): Future[Seq[ConsignmentRow]] = {
-    val query = Consignment.filterOpt(after)(_.consignmentreference > _)
+  def getConsignments(limit: Int, after: Option[String], userId: Option[UUID] = None): Future[Seq[ConsignmentRow]] = {
```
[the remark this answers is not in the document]
vimleshtna: Yeah, good point, updated it, thanks.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
vimleshtna: Actually, there is no check related to the filter in the `consignmentService.getConsignments` method. So, it is just a naming. I have changed it to `return filtered consignments when consignment filter is passed`.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
vimleshtna: Yeah, that makes sense. I have removed the method itself and passed the `consignmentFilters` to the service.

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/ConsignmentRouteSpec.scala
```diff
@@ -643,6 +643,115 @@ class ConsignmentRouteSpec extends TestContainerUtils with Matchers with TestReq
...
+
+      val userAccessToken = validUserTokenNoBody
+
+      val expectedResponse: GraphqlConsignmentsQueryData = expectedConsignmentsQueryResponse("data_userId")
+      val response: GraphqlConsignmentsQueryData = runConsignmentsTestQuery("query_with_userId", userAccessToken)
+
+      response.data.get.consignments should equal(expectedResponse.data.get.consignments)
+  }
+
+  "consignments" should "allow a user with reporting access can filter consignment by userId" in withContainers {
```
TomJKing, line 675: Should reword this test slightly: `... access to filter consignments by userId`

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/ConsignmentFields.scala
```diff
@@ -5,7 +5,8 @@ import io.circe.generic.auto._
 import sangria.macros.derive._
 import sangria.marshalling.circe._
 import sangria.relay._
-import sangria.schema.{Argument, BooleanType, Field, InputObjectType, IntType, ListType, ObjectType, OptionInputType, OptionType, ProjectedName, Projector, StringType, fields}
+import sangria.schema.{Argument, BooleanType, Context, Field, InputObjectType, IntType, ListType, ObjectType, OptionInputType, OptionType, Projector, StringType, fields}
+import uk.gov.nationalarchives.tdr.api.auth.ValidateHasConsignmentsAccess.reportingRole
```
TomJKing, line 9: Think this import is unused now: `uk.gov.nationalarchives.tdr.api.auth.ValidateHasConsignmentsAccess.reportingRole`

TomJKing (commented):
Looks good. Just a couple of extra minor comments. Thanks

TomJKing (approved):
Looks good. Thanks

PR #424, opened 2022-09-21
TDR 2298 add ordinal to api query

MancunianSam:
Add ordinal to API query
    
This is the ordinal on file property values

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,4 @@
 import rocks.muki.graphql.quietError
-import rocks.muki.graphql.schema.SchemaLoader
-import sbt.File
 import sbt.Keys.libraryDependencies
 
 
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -44,6 +44,11 @@ lazy val akkaHttpVersion = "10.2.10"
 lazy val circeVersion = "0.14.3"
 lazy val testContainersVersion = "0.40.10"
 
+dependencyOverrides ++= Seq(
```
TomJKing, line 47: This is in the `build.sbt` file twice. Also on line 89: https://github.com/nationalarchives/tdr-consignment-api/blob/master/build.sbt#L89. Probably should remove one of them

TomJKing (commented):
Looks good. One minor comment. Thanks

build.sbt
MancunianSam: I'm blind, I didn't notice it was already in there. I've removed the one I added.

TomJKing (approved):
Looks good. Thanks

PR #428, opened 2022-09-26
Changes for downloadable report.

MancunianSam:
I've added in a key value pair field in the standard consignment query. We'll eventually move everything over to use this but for now we'll run it in parallel with the existing `metadata` field.

I've also added the exportOrdinal and allowExport fields to the custom metadata output.


TomJKing (approved):
Looks good. If you haven't already could you add a JIRA task to move over to the new model for metadata properties. Thanks

PR #445, opened 2022-10-12
Update ClientChecks status.

MancunianSam:
I was adding the ClientChecks status row to the ConsignmentStatus table
on startUpload and I was trying to add it again when the metadata was
complete.

This doesn't work as there should only be one ConsignmentStatus row so
I've called updateConsignmentStatus.

I've added a test to check for this.


vimleshtna (approved):
Looks good, thanks.

PR #451, opened 2022-10-14
Allow getFileStatus to accept multiple statusTypes

techncl:


src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -152,7 +152,7 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
     when(fileRepositoryMock.getRedactedFilePairs(consignmentId)).thenReturn(Future(Seq()))
     when(ffidMetadataRepositoryMock.getFFIDMetadata(consignmentId)).thenReturn(Future(List()))
     when(antivirusRepositoryMock.getAntivirusMetadata(consignmentId)).thenReturn(Future(List()))
-    when(fileStatusRepositoryMock.getFileStatus(consignmentId, "FFID")).thenReturn(mockFileStatusResponse)
+    when(fileStatusRepositoryMock.getFileStatus(consignmentId, Set("FFID"))).thenReturn(mockFileStatusResponse)
```
MancunianSam, line 155: Can you use the `FFID` variable in `FileStatusService` instead of the String value here? You're already using it in some of the other tests.

MancunianSam (commented):
Thanks for doing this, it is a lot tidier now I think. Just one minor comment.

MancunianSam (approved):

PR #467, opened 2022-10-31
Temporarily check for file upload failures rather than successes

techncl:
updateConsignmentStatus is checking that all file Upload statuses are marked as 'Success', since we are setting them later on, there are 0 files with upload statuses, causing an exception to be thrown. This is a temporary fix that will just check if the file has failed, otherwise, it will label it as 'Completed'. It also comments out the tests that check for the no file exception.


TomJKing (approved):
Looks good. Thanks

PR #473, opened 2022-11-02
Add 'DescriptionClosed' to FileMetadata Object

techncl:
Even though it retrieves 'DescriptionPublic' (which has the opposite meaning), we are planning to rename 'DescriptionPublic' in the future


TomJKing (approved):
Looks good. Thanks

PR #476, opened 2022-11-04
Rename descriptionPublic to DescriptionClosed part 1

thanhz:
This is a temporary change so that the migrations don't break the existing form. There will be another pr to use only "DescriptionClosed" once the migrations are in.


techncl (approved):
Looks good!

PR #480, opened 2022-11-07
Increase some timeouts.

MancunianSam:
This is part of TDR-2509. Uploads with large numbers of files are timing
out but the database isn't running at capacity so we just need to wait a
bit longer for results.

I've tried these timeouts with 10000 files and it's working.


TomJKing (approved):
Looks good. Thanks

PR #481, opened 2022-11-10
Add Filename and FileType to the metadata.

MancunianSam:
This happens when the file is added along with all the other metadata
with a default value.

We already have a few tests which are checking the right number of
metadata rows is inserted so I've updated these.


TomJKing (approved):
Looks good. Thanks

PR #482, opened 2022-11-11
Tdr 2477 simplify updating metadata

TomJKing:


vimleshtna (commented):
Now, it looks very clean and clear. Thanks Tom. Just a minor suggestion.
Also, can we delete the method `updateFileMetadataProperties` as we are not using this anymore https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala#L62

vimleshtna (approved):
Looks good, thanks

PR #485, opened 2022-11-15
Update redacted files logic

MancunianSam:
It turns out that redacted and original files can have different file
extensions so I've updated the SQL query.

Instead of converting file1_R.txt to file1.txt and comparing it with
file1.txt, it now converts file1_R.txt to file1. and compares it with
file1. which will match missing extensions.

It will now filter out any original files that don't have a suffix as
well.


TomJKing (approved):
Looks good. Thanks.

Was there a reason the regEx was done in the SQL? Just curious.

MancunianSam:
@TomJKing I don't think there was a specific reason. We could have bought all the rows back and done the regex in the code but this reduces the rows returned from the DB.

PR #486, opened 2022-11-15
Use DescriptionClosed instead of DescriptionPublic

thanhz:


thanhz:
> Looks good. There is just one instance in the "FileServiceSpec" where there is still "DescriptionPublic" which would get confusing later on.

@techncl hmm it might just be my machine playing up again, but I can't seem to find the "DescriptionPublic" in the "FileServiceSpec" >.<

techncl (approved):
Never mind; I was looking at the wrong branch 🤦

PR #493, opened 2022-11-28
Fix for the redacted files.

MancunianSam:
The underscore character is a wildcard character in postgres
https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP

The similar to is then looking for anything which ends with R.extension
instead of looking for something that ends in _R.extension. The
filenames which were failing all ended in R.

I've added another test to check for this which I think should cover it.


TomJKing (approved):
Looks good. Thanks

PR #494, opened 2022-11-29
Add bulk file metadata updates

MancunianSam:
This is part of the change to the backend checks architecture. We're now going to be updating the API once all the files are processed. I'm going to keep the FFID, AV and Checksum updates separate because I don't want to overload the database too much.
I've left the original addFFIDMetadata in so the existing FFID checks will still work but once they're moved over, we can delete it.
I've updated all the tests to point to the new endpoint.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
```diff
@@ -27,37 +26,42 @@ class FFIDMetadataService(
...
-      uuidSource.uuid,
-      ffidMetadata.fileId,
-      ffidMetadata.software,
-      ffidMetadata.softwareVersion,
-      Timestamp.from(timeSource.now),
-      ffidMetadata.binarySignatureFileVersion,
-      ffidMetadata.containerSignatureFileVersion,
-      ffidMetadata.method
-    )
+  @deprecated("Use addFFIDMetadata(ffidMetadataInput: FFIDMetadataInput)")
```
TomJKing, line 29: Nice.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
```diff
@@ -27,37 +26,42 @@ class FFIDMetadataService(
...
+      })
+      .unzip
 
-    (for {
-      fileStatusRows <- generateFileStatusRows(ffidMetadata)
-      _ = loggingUtils.logFileFormatStatus("FFID", ffidMetadata.fileId, fileStatusRows.map(_.value).mkString(","))
-      ffidMetadataRow <- ffidMetadataRepository.addFFIDMetadata(metadataRow, fileStatusRows)
-      ffidMetadataMatchesRow <- addFFIDMetadataMatches(ffidMetadataRow.ffidmetadataid)
+    for {
+      statuses <- Future.sequence(ffidMetadataInput.metadataInputValues.map(generateFileStatusRows))
```
TomJKing, line 59: Is there a reason that we're no longer logging the statuses?

TomJKing (commented):
Looks good. Just a couple of very minor comments. Thanks

For my own sanity this won't be a breaking change will it?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FFIDMetadataService.scala
MancunianSam: I think I had some vague idea about replacing the Grafana monitoring that runs off these logs with something in the step functions but that's way off so I've put the logging back.

TomJKing (approved):
Looks good. Thanks

PR #495, opened 2022-11-29
Add bulk AV metadata mutation

MancunianSam:
This is part of the change to the backend checks architecture. We're now going to be updating the API once all the files are processed. I'm going to keep the FFID, AV and Checksum updates separate because I don't want to overload the database too much.
I've left the original addAntivirusMetadata in so the existing AV checks will still work but once they're moved over, we can delete it.
I've updated all the tests to point to the new endpoint.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/AntivirusMetadataService.scala
```diff
@@ -1,55 +1,57 @@
...
-      case _  => VirusDetected
-    }
-    val fileStatusRow = FilestatusRow(uuidSource.uuid, input.fileId, Antivirus, fileStatusValue, Timestamp.from(timeSource.now))
-    val avMatch = if (fileStatusValue == VirusDetected) { s": ${input.result}" }
-    else { "" }
-
-    loggingUtils.logFileFormatStatus("antivirus", input.fileId, fileStatusValue + avMatch)
-
-    antivirusMetadataRepository.addAntivirusMetadata(inputRow, fileStatusRow).map(rowToAntivirusMetadata)
+  @deprecated("Use addAntivirusMetadata(input: AddAntivirusMetadataInput): Future[List[AntivirusMetadata]] instead")
```
TomJKing, line 17: Nice, we really should be using the deprecated flag a lot more in the API code

src/main/scala/uk/gov/nationalarchives/tdr/api/service/AntivirusMetadataService.scala
```diff
@@ -1,55 +1,57 @@
...
+      .map(values => {
+        val inputRow =
+          AvmetadataRow(values.fileId, values.software, values.softwareVersion, values.databaseVersion, values.result, Timestamp.from(Instant.ofEpochMilli(values.datetime)))
+        val fileStatusValue = values.result match {
+          case "" => Success
+          case _  => VirusDetected
+        }
+        val fileStatusRow = FilestatusRow(uuidSource.uuid, values.fileId, Antivirus, fileStatusValue, Timestamp.from(timeSource.now))
+        (inputRow, fileStatusRow)
+
```
TomJKing, line 32: Unnecessary line break

TomJKing (commented):
Looks good. Just a couple of very minor comments.

For my own sanity, this won't cause any clients calling this endpoint to break?

MancunianSam:
The existing endpoint will still work. The name of the input case class has changed but the fields in it are the same.

TomJKing (approved):
Looks good. Thanks

PR #496, opened 2022-11-30
Add multiple file metadata updates

MancunianSam:
Create a new multiple file metadata endpoint.
    
The existing updateBulkFileMetadata adds the same metadata values to multiple files. For the new backend checks, we need a way of adding different property values (ServerSideChecksum) to multiple files.
    
I've added this in and updated the tests to point to the new one. It's only the api update lambda that uses this endpoint so once this is all deployed, we can remove the old one.


schema.graphql
```diff
@@ -19,6 +19,10 @@ input AddFileAndMetadataInput {
 }
 
 input AddFileMetadataWithFileIdInput {
+  metadataInputValues: [AddFileMetadataWithFileIdInputValues!]!
+}
+
+input AddFileMetadataWithFileIdInputValues {
```
TomJKing, line 25: Maybe rename this `ÀddFileMetadataInputValues` the "FileId" seems a bit superfluous.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileMetadataFields.scala
```diff
@@ -24,7 +24,9 @@ object FileMetadataFields {
 
   case class BulkFileMetadata(fileIds: Seq[UUID], metadataProperties: Seq[FileMetadata])
 
-  case class AddFileMetadataWithFileIdInput(filePropertyName: String, fileId: UUID, value: String) extends FileMetadataBase
+  case class AddFileMetadataWithFileIdInput(metadataInputValues: List[AddFileMetadataWithFileIdInputValues])
+
+  case class AddFileMetadataWithFileIdInputValues(filePropertyName: String, fileId: UUID, value: String) extends FileMetadataBase
```
TomJKing, line 29: See comment in schema about renaming this.

TomJKing (commented):
Looks good. Just a couple of minor comments.

schema.graphql
MancunianSam: I've updated the case class which will update this input as well. I think this name change will break the API update so we'll have to think about how to deploy this.

TomJKing (approved):
Looks good. Thanks

PR #502, opened 2022-12-08
Update delete metadata endpoint to take properties

techncl:


TomJKing (commented):
I'm wondering if this can be simplified a bit, by making use of the `CustomMetadataService`, as this does the heavy lifting already, in terms of getting a property's dependencies, and default values. 

Maybe something like this:

```
def deleteFileMetadata (input: DeleteFileMetadataInput, userId: UUID): Future[DeleteFileMetadata] = {
    for {
      customMetadataProperties <- customMetadataService.getCustomMetadata
    }
     yield {
      val propertyNames = input.propertyNamesAndValues.map(_.filePropertyName)
      val propertiesToReset: Seq[CustomMetadataField] = customMetadataProperties.filter(cmp => propertyNames.contains(cmp.name))
      val allPropertiesToDelete: Set[String] = propertiesToReset.flatMap(p => {
        val dependencies: List[String] = p.values.flatten(_.dependencies.map(_.name))
        dependencies :+ p.name
      }).toSet
       
      val propertyDefaults: Seq[(String, String)] = customMetadataProperties.filter(cmp => allPropertiesToDelete.contains(cmp.name)).map(cmp => {
        cmp.defaultValue match {
          case Some(value) => (cmp.name, value)
        }
      })

      val resetMetadata = input.fileIds.flatMap(id => {
        propertyDefaults.map(r => {
          FilemetadataRow(uuidSource.uuid, id, r._2, Timestamp.from(timeSource.now), userId, r._2)
        })
      })

      fileMetadataRepository.deleteFileMetadata(input.fileIds.toSet, allPropertiesToDelete)
      fileMetadataRepository.addFileMetadata(resetMetadata)

      DeleteFileMetadata(input.fileIds, allPropertiesToDelete.toSeq)
    }
  }
```
You can then just pass in the property names, file ids and user id, without the need for the value. You can use the response from the `CustomMetadataService` to work out everything else. 

What do you think?

techncl:
That's a good idea; I completely forgot about that Service. Cool, I'll do that. Also, yes, we probably don't need to pass in the "value", at least for the type of fields we have now.

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileMetadataFields.scala
```diff
@@ -32,7 +34,10 @@ object FileMetadataFields {
 
   case class DeleteFileMetadata(fileIds: Seq[UUID], filePropertyNames: Seq[String])
 
-  case class DeleteFileMetadataInput(fileIds: Seq[UUID])
+  case class DeleteFileMetadataInput(
```
TomJKing, line 37: Is adding the additional field `propertyNames` not going to break any client using this input?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -83,43 +81,80 @@ class FileMetadataService(
...
+      namesOfAllPropertiesToDelete: Set[String] = customMetadataProperties
+        .collect {
+          case customMetadataProperty if input.propertyNames.contains(customMetadataProperty.name) =>
+            val namesOfDependenciesToDelete: List[String] = customMetadataProperty.values.flatMap(_.dependencies.map(_.name))
+            namesOfDependenciesToDelete :+ customMetadataProperty.name
+        }
+        .flatten
+        .toSet
+
+      _ = if (namesOfAllPropertiesToDelete.isEmpty) {
```
TomJKing, line 98: Don't think there is any need to throw an exception, if there is nothing to delete there is nothing to delete.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -83,43 +81,80 @@ class FileMetadataService(
...
+        }
+      }.toSeq
+      _ <- fileMetadataRepository.deleteFileMetadata(fileIds, namesOfAllPropertiesToDelete)
+      _ <- fileMetadataRepository.addFileMetadata(metadataToReset)
+    } yield DeleteFileMetadata(fileIds.toSeq, namesOfAllPropertiesToDelete.toSeq)
+  }
 
-    val fileMetadataUpdates =
-      fileMetadataToUpdate.map(propertyName => propertyName -> update(propertyName, metadataValuesWithDefault.find(_.propertyname == propertyName).head.propertyvalue)).toMap
+  private def getFileMetadataToDeleteAndReset(
```
TomJKing, line 119: You don't need this method any more

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -245,5 +280,7 @@ object FileMetadataService {
 
   case class FileMetadataDelete(fileIds: Set[UUID], propertyNamesToDelete: Set[String])
 
+  case class FileMetadataValueDeletions(valuesToReset: Seq[FileMetadataUpdate], valuesToDelete: Seq[String])
```
TomJKing, line 283: You don't need this case class anymore

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataServiceSpec.scala
```diff
@@ -310,57 +310,143 @@ class FileMetadataServiceSpec extends AnyFlatSpec with MockitoSugar with Matcher
...
+    }
 
+    verify(fileMetadataRepositoryMock, times(0)).addFileMetadata(any[Seq[FilemetadataRow]])
+    verify(fileMetadataRepositoryMock, times(0)).deleteFileMetadata(any[Set[UUID]], any[Set[String]])
+    verify(customMetadataPropertiesRepositoryMock, times(1)).getCustomMetadataDependencies
+
+    thrownException.getMessage should include("Can't find metadata property 'Non-ExistentProperty' in the db.")
+  }
+
+  "deleteFileMetadata" should "be able to handle requests to delete and/or reset multiple top-level properties (whether they have dependencies or not)" in {
```
TomJKing, line 353: Maybe rename this test slightly: `should "delete and/or reset multiple properties including any dependencies"`

TomJKing (commented):
Looks good. Just some minor comments. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileMetadataFields.scala
techncl: Hmm, it shouldn't due to the default argument ` = Seq(ClosureType)`. I added [a card](https://national-archives.atlassian.net/browse/TDR-2723) to then add the argument explicitly

TomJKing: OK.

TomJKing (approved):
Looks good. Thanks

PR #504, opened 2022-12-12
Tdr 2693 display properties

TomJKing:


vimleshtna (commented):
Looks good, just a minor comment.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.nationalarchives.tdr.api.db.repository.DisplayPropertiesRepository
+import uk.gov.nationalarchives.tdr.api.graphql.fields.DisplayPropertiesFields._
+import uk.gov.nationalarchives.tdr.api.graphql.fields.DisplayPropertiesFields
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DisplayPropertiesService(displayPropertiesRepository: DisplayPropertiesRepository)(implicit val ec: ExecutionContext) {
+
+  implicit class DisplayPropertyRowHelper(properties: Seq[Tables.DisplaypropertiesRow]) {
+    def toDisplayProperty: Seq[DisplayPropertyField] = {
```
techncl, line 13: Does this need to be an implicit as it's only being used once? Couldn't it be a private method that display properties gets passed into?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
```diff
@@ -0,0 +1,39 @@
...
+      properties
+        .groupBy(_.propertyname)
+        .map(r => {
```
techncl, line 16: Just two suggestions, feel free to ignore:

1. If you replace the brackets of the `map`s with braces, you can get rid of the brace after the `=>` which would make it cleaner

```
    def toDisplayProperty: Seq[DisplayPropertyField] = {
      properties
        .groupBy(_.propertyname)
        .map{r =>
          val name = r._1.get
          val attributes = r._2.map{a =>
            val attribute: String = a.attribute.getOrElse(
              throw new Exception(
                s"Error: Property name '$name' has empty attribute name"
              )
            )
            DisplayAttribute(attribute, a.value, DisplayPropertiesFields.toDataType(a.attributetype))
          }
          DisplayPropertyField(name, attributes)
        }
        .toSeq
    }
```


2. If you decided to use braces (instead of brackets) in the `map`, you can add `case ({nameOfFirstElement}, {nameOfSecondElement})` allowing you to name the elements, instead of `r._1` and `r._2`
```
    def toDisplayProperty: Seq[DisplayPropertyField] = {
      properties
        .groupBy(_.propertyname)
        .map { case (displayName, displayProperties) =>
          val name = displayName.get
          val attributes = displayProperties.map { a =>
            val attribute: String = a.attribute.getOrElse(
              throw new Exception(
                s"Error: Property name '$name' has empty attribute name"
              )
            )
            DisplayAttribute(attribute, a.value, DisplayPropertiesFields.toDataType(a.attributetype))
          }
          DisplayPropertyField(name, attributes)
        }
        .toSeq
    }
```

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/DisplayPropertiesRouteSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+import uk.gov.nationalarchives.tdr.api.utils.{TestContainerUtils, TestRequest, TestUtils}
+
+import java.util.UUID
+
+class DisplayPropertiesRouteSpec extends TestContainerUtils with Matchers with TestRequest {
+  sealed trait DataType
+  case object Text extends DataType
+  case object Integer extends DataType
+  case object DateTime extends DataType
+  case object Decimal extends DataType
```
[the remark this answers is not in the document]
TomJKing: Removed them.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
TomJKing: Fair point, getting a bit carried away with implicits

vimleshtna (dismissed):
Looks good, thanks

techncl (approved):
Looks good!

techncl (approved):

PR #510, opened 2022-12-22
Ensure file metadata in correct state if 'description' deleted

TomJKing:
Cannot have a 'DescriptionAlternate' property without a 'description' property

'DescriptionAlternate' property is a dependency of 'DescriptionClosed' property

If 'description' is deleted then 'DescriptionClosed' property to be set back to default of 'false' and 'DescriptionAlternate' to be deleted


vimleshtna (approved):
Looks goos, thanks

PR #514, opened 2022-12-30
Tdr 2760 delete empty properties

TomJKing:


vimleshtna (commented):
I was thinking shouldn't we handle the delete part on the client side?
Because this is an update call and we should do an update only.
We can call delete API from the client side for the empty values.
We are already calling delete API for the TitleClosed and DescriptionClosed dependencies, we may utilize this for the description field

What do you think about this approach?

TomJKing:
> _I was thinking shouldn't we handle the delete part on the client side?_
@vimleshtna considered doing this as well. My feeling is this should be down to the API, to ensure the DB doesn't get into an invalid state. The API shouldn't rely on any clients using it correctly. Whilst at the moment the risk is lower, the more properties/dependencies added in, the more likely a client is going to misuse the API and put invalid data into the DB.

vimleshtna:
> > _I was thinking shouldn't we handle the delete part on the client side?_
> 
> @vimleshtna considered doing this as well. My feeling is this should be down to the API, to ensure the DB doesn't get into an invalid state. The API shouldn't rely on any clients using it correctly. Whilst at the moment the risk is lower, the more properties/dependencies added in, the more likely a client is going to misuse the API and put invalid data into the DB.

@TomJKing Yeah, you are right. I think we should change the endpoint `updateBulkFileMetadata` to updateOrDeleteBulkFileMetadata or throw an exception if the user passes an empty value which will ensure that the user has to pass the correct value if he is calling an update API.

TomJKing:
> I think we should change the endpoint updateBulkFileMetadata to updateOrDeleteBulkFileMetadata or throw an exception if the user passes an empty value
@vimleshtna good point, I've changed the endpoint to throw an exception if an empty property value is passed in. That should prevent misuse by any clients.

vimleshtna (approved):
Looks good, thanks.

PR #518, opened 2023-01-04
Add additionalProperties in metadata filter so that we can get other …

vimleshtna:
…properties too


schema.graphql
```diff
@@ -327,6 +327,7 @@ type FileMetadata {
 input FileMetadataFilters {
   closureMetadata: Boolean = false
   descriptiveMetadata: Boolean = false
+  additionalProperties: [String!]
```
TomJKing, line 330: I would rename this to simply `properties` as its just a filter of the property names.

TomJKing (approved):
Looks good. Thanks

PR #520, opened 2023-01-06
TDR-2494-Remove finalOpenRecordsConfirmed

thanhz:
Removal of finalOpenRecordsConfirmed property so that we can remove the open records checkbox on the confirmTransfer page.


TomJKing (commented):
Looks good. Could you make sure that there is a PR ready in the e2e tests to take account of this change? I'm not sure if it will affect the e2e tests, but I think it will.

PR #528, opened 2023-01-10
Add export exception to AuthorisationTag

MancunianSam:
This affects the addMultipleFileStatuses mutation. We're calling this
from the backend checks to add file statuses so it needs an exemption to
the "must own your files rule" This may be called by users tokens in the
future though so I don't want to make this a "Backend checks only"
endpoint.


TomJKing (commented):
Looks good. Just one minor comment

PR #529, opened 2023-01-10
Remove default 'ClosureType' argument from deleteFileMetadata endpoint

techncl:


TomJKing (commented):
Looks good. Thanks. Have this been checked to make sure it doesn't break any clients using this endpoint?

TomJKing (approved):
Looks good. Thanks.

PR #535, opened 2023-01-16
Add client checks status to current status

techncl:


MancunianSam (approved):
Looks good. I like the test you've added.

PR #536, opened 2023-01-17
Remove file status inserts

MancunianSam:
We were adding all of the file and consignment status updates for the backend checks results in the API.
This has now been moved to the https://github.com/nationalarchives/tdr-statuses/ lambda function which then calls API update.

I've updated some of the tests and removed others where they're no longer relevant.

I've removed AllowedPuids and Disallowed puids as these are read by the tdr-statuses lambda.

This can't be deployed until the new backend checks are deployed. This will require downtime.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
```diff
@@ -1,13 +1,12 @@
 package uk.gov.nationalarchives.tdr.api.service
```
TomJKing, line 1: Realise its not your code changes but there are `validStatusTypes` and `validStatusValues` vals in the object, could you add a couple of unit tests to make sure these vals are checked? These are pretty crucial so should at least have some testing, until they get moved to a DB table.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -171,6 +171,7 @@ object FileMetadataService {
 
   val SHA256ClientSideChecksum = "SHA256ClientSideChecksum"
```
TomJKing, line 172: Realise its not your code, but could you add some unit tests to check all these hardcoded values. They are pretty crucial, and some test would be good to at least to make devs think before changing them.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
```diff
@@ -55,10 +48,9 @@ object FileStatusService {
   val ChecksumMatch = "ChecksumMatch"
   val Antivirus = "Antivirus"
   val FFID = "FFID"
+  val Redaction = "Redaction"
```
TomJKing, line 51: Could you update the unit tests to check this value, and update for the other values, as the unit tests on line 236, and 242 look out of date, as missing a few values.

TomJKing (commented):
Looks good (provided it works and doesn't break anything! ;) ) Just some very minor comments.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
MancunianSam: There's a couple of tests here https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusServiceSpec.scala#L628 which check the types and values are what we're expecting, is that enough?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
MancunianSam: I've updated the unit tests in `FileStatusServiceSpec` to check all of the values.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: I've added a test in to check this value. The only place it's being used is checking the status of the checksums which is pretty worthless not we're not updating one file at a time but I think I've changed enough for one PR.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -173,30 +163,25 @@ class FileServiceSpec extends AnyFlatSpec with MockitoSugar with Matchers with S
 
     when(fileRepositoryMock.getFiles(consignmentId, FileFilters(None)))
       .thenReturn(Future(fileAndMetadataRows))
-    when(fileRepositoryMock.getRedactedFilePairs(consignmentId)).thenReturn(Future(Seq()))
+//    when(fileRepositoryMock.getRedactedFilePairs(consignmentId)).thenReturn(Future(Seq()))
```
TomJKing, line 166: Can this code be deleted now?

TomJKing (commented):
Looks good. Just one last very minor comment.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
MancunianSam: Good spot, yeah, it can, I've removed it.

TomJKing (approved):
Looks good. Thanks

PR #545, opened 2023-01-25
Tdr 2254 custom metadata validation

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ValidateFileMetadataService.scala
```diff
@@ -0,0 +1,113 @@
...
+        List()
+      } else {
+        val additionalMetadataStatuses = {
+          additionalMetadataFieldGroups
+            .flatMap(group => {
+              val states = group.fields.flatMap(f => checkPropertyState(fileIds, f, existingMetadataProperties))
+              val filesWithNoAdditionalMetadataStatuses = fileIds
+                .filter(id => !states.map(_.fileId).contains(id))
+                .map(id => {
+                  FilestatusRow(UUID.randomUUID(), id, group.groupName, NotEntered, Timestamp.from(timeSource.now))
```
MancunianSam, line 52: This doesn't really matter because it's only there for the tests really but the other services have a `UUIDSource` passed in the constructor and they use that to get a new UUID. There's another `randomUUID` on line 64. Feel free to leave it if it causes problems though.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ValidateFileMetadataService.scala
```diff
@@ -0,0 +1,113 @@
...
+                })
+
+              val statuses = states
+                .groupBy(_.fileId)
+                .map(s => {
+                  val fileId = s._1
```
MancunianSam, line 58: Minor thing to maybe make it a bit tidier but you can extract both sides of the tuple and use those
```scala
                  val (fileId, states) = s
                  val status: String = states match {
                    case s if s.forall...
```
Not sure if it's better but it gets rid of the ._1 ._2 stuff.

MancunianSam (commented):
Looks good, thanks. Just a couple of suggestions.

MancunianSam (approved):

PR #547, opened 2023-01-30
TDR 2826 make all english open records optional

MancunianSam:
- Make initialOpenRecords and allEnglish optional.
- Update the tests and add two new ones


TomJKing (approved):
Looks good. Thanks

PR #549, opened 2023-01-31
Sort consignments, returned from getConsignments, in descending order

techncl:
In addition to sorting by descending order, the sorting was happening after the 'after' filter which seemed wrong; it's now happening before. This fix also means that it will no longer return values if an empty string was passed instead of a consignment reference, which seems to makes sense as we don't have empty strings as consignment reference values and if you don't want to filter by reference, then you can pass it a 'None.


TomJKing (approved):
Looks good. Thanks

PR #555, opened 2023-02-06
TDR-2755 - Return all file statuses

vimleshtna:
Return ClosureMetadata and DescriptiveMetadata file statuses too


TomJKing (approved):
Looks good. Thanks

PR #575, opened 2023-03-06
TDR-2658 Update the API to add metadata consignment statuses.

MancunianSam:
This now:
* Adds a DescriptiveMetadata and ClosureMetadata status when the consignment is created
* Updates each status based on the value of the validated file statuses.

I've updated/added tests where necessary.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -70,7 +68,9 @@ class FileMetadataService(
       fileIds: Set[UUID] = existingFileRows.toFileTypeIds
       _ <- fileMetadataRepository.deleteFileMetadata(fileIds, distinctPropertyNames)
       addedRows <- fileMetadataRepository.addFileMetadata(generateFileMetadataRows(fileIds, distinctMetadataProperties, userId))
-      _ <- validateFileMetadataService.validateAdditionalMetadata(uniqueFileIds.toSet, input.consignmentId, distinctPropertyNames)
+      fileStatuses <- validateFileMetadataService.validateAdditionalMetadata(uniqueFileIds.toSet, input.consignmentId, distinctPropertyNames)
+      _ <- consignmentStatusService.updateMetadataConsignmentStatus(consignmentId, fileStatuses, DescriptiveMetadata)
```
TomJKing, line 72: Could this update be done in a single call, with the `updateMetadataConsignmentStatus` function taking a list of types?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -108,7 +108,10 @@ class FileMetadataService(
       }.toSeq
       _ <- fileMetadataRepository.deleteFileMetadata(fileIds, allPropertiesToDelete)
       _ <- fileMetadataRepository.addFileMetadata(metadataToReset)
-      _ <- validateFileMetadataService.validateAdditionalMetadata(fileIds, existingFileRows.map(_.consignmentid).head, allPropertiesToDelete)
+      fileStatuses <- validateFileMetadataService.validateAdditionalMetadata(fileIds, existingFileRows.map(_.consignmentid).head, allPropertiesToDelete)
+      consignmentId = existingFileRows.map(_.consignmentid).headOption.getOrElse(throw InputDataException("No consignment id found for files"))
+      _ <- consignmentStatusService.updateMetadataConsignmentStatus(consignmentId, fileStatuses, DescriptiveMetadata)
```
TomJKing, line 113: See comment above, about making a single call

src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala
```diff
@@ -91,17 +95,47 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
...
+
+    val sortedValues = rowCaptor.getAllValues.asScala.sortBy(r => r.statustype)
+    sortedValues.head.consignmentid should be(result.consignmentid)
+    sortedValues.head.statustype should be(ClosureMetadata)
+    sortedValues.head.value should be(NotEntered)
+
+    sortedValues.last.consignmentid should be(result.consignmentid)
+    sortedValues.last.statustype should be(DescriptiveMetadata)
+    sortedValues.last.value should be(NotEntered)
+
```
TomJKing, line 124: Unnecessary line break

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
```diff
@@ -91,7 +105,7 @@ class ConsignmentStatusService(
 }
 
 object ConsignmentStatusService {
-  val validConsignmentTypes: List[String] = List("Series", "TransferAgreement", "Upload", "ClientChecks", "ConfirmTransfer", "Export")
+  val validConsignmentTypes: List[String] = List("Series", "TransferAgreement", "Upload", "ClientChecks", "ClosureMetadata", "DescriptiveMetadata", "ConfirmTransfer", "Export")
```
TomJKing, line 108: I'm going to raise a new JIRA ticket about consolidating the consignment and file status types /values in to a set of case objects, instead of hard coding these strings in multiple places in the code.

TomJKing (commented):
Looks good. Couple of minor comments. Once the `NotEntered` statuses are being added to all new consignments, we should run a DB migration to backfill the old ones, for consistency.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: Yeah, that's a good point. I've changed it to take a list of status types and removed the second call.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -70,7 +68,8 @@ class FileMetadataService(
       fileIds: Set[UUID] = existingFileRows.toFileTypeIds
       _ <- fileMetadataRepository.deleteFileMetadata(fileIds, distinctPropertyNames)
       addedRows <- fileMetadataRepository.addFileMetadata(generateFileMetadataRows(fileIds, distinctMetadataProperties, userId))
-      _ <- validateFileMetadataService.validateAdditionalMetadata(uniqueFileIds.toSet, input.consignmentId, distinctPropertyNames)
+      fileStatuses <- validateFileMetadataService.validateAdditionalMetadata(uniqueFileIds.toSet, input.consignmentId, distinctPropertyNames)
+      _ <- consignmentStatusService.updateMetadataConsignmentStatus(consignmentId, fileStatuses, List(DescriptiveMetadata, ClosureMetadata))
```
vimleshtna, line 72: I'm not sure if you already considered the following scenarios but it's worth adding unit tests:

Update:

1. Set `Incomplete` consignment status if one of the existing file's status is `Incomplete`
2. Set `Complete` consignment status if none of the existing files' statuses are `Incomplete`

Delete:
1. Set `Incomplete` consignment status if one of the existing file's status is `Incomplete`
2. Set `Complete` consignment status if one of the existing file's status is `Complete`
2. Set `NotEntered` consignment status if all of the existing files' statuses are `NotEntered`

One more thing, since we are not going to add a database trigger at the moment, I was thinking to do the update consignment job asynchronously because the update file metadata call is a bit slow as we are validating on the frontend and API too, and if we do the consignment status update task (which is not related to the update/delete action) then it will be very heavy considering if we have thousands of files for a consignment.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -71,17 +72,22 @@ class ConsignmentService(
...
         consignmentsequence = sequence,
         consignmentreference = consignmentRef,
         consignmenttype = consignmentType,
         bodyid = body.bodyId
       )
+      descriptiveMetadataStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, DescriptiveMetadata, "NotEntered", timestampNow, Option(timestampNow))
+      closureMetadataStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, ClosureMetadata, "NotEntered", timestampNow, Option(timestampNow))
       consignment <- consignmentRepository.addConsignment(consignmentRow).map(row => convertRowToConsignment(row))
+      _ <- consignmentStatusRepository.addConsignmentStatus(descriptiveMetadataStatusRow)
+      _ <- consignmentStatusRepository.addConsignmentStatus(closureMetadataStatusRow)
```
vimleshtna, line 90: We may create another method `addConsignmentStatuses` which will accept statuses as a list and pass `descriptiveMetadataStatusRow` and `closureMetadataStatusRow` as a list to do the transition in a single call.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -71,17 +72,22 @@ class ConsignmentService(
...
         addConsignmentInput.seriesid,
         token.userId,
-        Timestamp.from(now),
+        timestampNow,
         consignmentsequence = sequence,
         consignmentreference = consignmentRef,
         consignmenttype = consignmentType,
         bodyid = body.bodyId
       )
+      descriptiveMetadataStatusRow = ConsignmentstatusRow(uuidSource.uuid, consignmentId, DescriptiveMetadata, "NotEntered", timestampNow, Option(timestampNow))
```
vimleshtna, line 86: We may use the variable for the "NotEntered" string.

vimleshtna (commented):
Looks good to me, just highlighting a couple of points, it's worth checking them, thank you.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
MancunianSam: Those scenarios are [covered in the unit tests](https://github.com/nationalarchives/tdr-consignment-api/pull/575/files#diff-8af430ff536b04425b75b9adbd9c69938ea74433c1230ab729656c72206e0f29R393) for the `updateMetadataConsignmentStatus` method  
I don't think there's much point in checking them again here. The unit tests for this method in `FileMetadataService` check that the `updateMetadataConsignmentStatus` which is enough I think. 

For the second point, it depends what you mean by asynchronously. If you mean taking the `Future` out of the for comprehension then that's not a good idea as it can lead to some odd behaviour down the line. There are things we can do with Akka actors and bits but we're trying to get rid of Akka so I don't want to get more embedded in their ecosystem. I'm not sure how else we'd do it.
The other problem with doing this asynchronously is that there is no way for the frontend to know if there's a failure so if the insert fails, the data will be inconsistent and the user will see the wrong status.

MancunianSam: I've added in tests for the add and delete metadata which check which status it returns if there is an existing file outside of the update with a particular status.
I've also changed the code so the tests work.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
MancunianSam: Good idea, I've added the method in and a repository test for it.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
```diff
@@ -78,6 +81,27 @@ class ConsignmentStatusService(
...
+            } else {
+              Completed
+            }
+            val input = ConsignmentStatusInput(consignmentId, statusType, Option(statusValue))
+            updateConsignmentStatus(input)
+          })
+          .sequence
+      })
+  }
+
```
vimleshtna, line 104: I just realized, if we get the correct consignment status from the SQL query itself, it might be a bit faster.
What do you think?

I tried to create one but I haven't tested it with a different set of data:

```

SELECT "StatusType", (CASE WHEN COUNT(CASE WHEN "Value" = 'Incomplete' THEN 1 ELSE NULL END) > 0 THEN 'Incomplete'
						   ELSE (CASE WHEN COUNT(CASE WHEN "Value" = 'Completed' THEN 1 ELSE NULL END) > 0 THEN 'Completed' ELSE 'NotEntered' END)
						   END) as Value						   						   
FROM "FileStatus" fs JOIN "File" f ON f."FileId" = fs."FileId"
WHERE f."ConsignmentId" = 'b7cb69b7-3df3-4e00-886b-1261b942d6ff'
GROUP BY "StatusType" HAVING "StatusType" IN ('ClosureMetadata', 'DescriptiveMetadata');
```

MancunianSam: I think it might be best to leave this for now. We've had raw SQL before and it has caused problems and I'm not sure if this will be too much faster. If there are problems that come out of the load testing then we can re-visit this.

vimleshtna (approved):

PR #580, opened 2023-03-08
Use file statuses for checks count.

MancunianSam:
At the moment, we're checking the relevant metadata fields to see if the
checks have completed. The problem is that if they fail, there is no
metadata so the checks page just hangs.

This checks the file statuses instead of the metadat itself because
these will be set to failed if there are any failures.


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -125,14 +125,6 @@ class ConsignmentService(
     consignmentRepository.consignmentHasFiles(consignmentId)
```
TomJKing, line 125: Looks like there are some unused constructor params now the method is remove, these should be removed as well:

- `fileMetadataRepository: FileMetadataRepository`
- `fileRepository: FileRepository`
- `ffidMetadataRepository: FFIDMetadataRepository`

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
```diff
@@ -25,6 +26,19 @@ class FileStatusService(fileStatusRepository: FileStatusRepository, uuidSource:
     fileStatusRepository.addFileStatuses(rows).map(_.map(row => FileStatus(row.fileid, row.statustype, row.value)).toList)
   }
 
+  def getConsignmentFileProgress(consignmentId: UUID): Future[FileChecks] = {
```
TomJKing, line 29: Should add some unit tests for this new function

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -125,14 +125,6 @@ class ConsignmentService(
...
   }
 
-  def getConsignmentFileProgress(consignmentId: UUID): Future[FileChecks] = {
```
TomJKing, line 128: I think the following functions can now be deleted, as they don't appear to be used anywhere expect in their own unit tests, and the unit tests can be removed as well:

- `def countProcessedAvMetadataInConsignment(consignmentId: UUID): Future[Int]`
- `def countProcessedChecksumInConsignment(consignmentId: UUID): Future[Int]`
- `def countProcessedFfidMetadata(consignmentId: UUID): Future[Int]`

TomJKing (commented):
Looks good. Just some minor comments. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala
MancunianSam: I've added some in, thanks.

MancunianSam: I've added some in, thanks.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
MancunianSam: I can't believe I missed a chance to go deleting large parts of the API code. I've removed them and their associated tests.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
MancunianSam: I've taken these out as well. The service is looking much tidier now.

TomJKing (approved):
Looks good. Thanks

PR #586, opened 2023-03-17
Wait for metadata to be added

MancunianSam:
Wait for metadata to be added.
    
`deleteFileStatus` and `addFileStatuses` were not flatMapped so sometimes they weren't completing before we were reading the file statuses to  determine the consignment statuses which lead to wrong results.
    
This adds these two methods into the for comprehension and this seems to be ok now.

I've also updated a couple of tests to mock out the calls to the methods.


TomJKing (approved):
Looks good. Thanks.

PR #593, opened 2023-03-29
Add consignment filter while calculating hte total number of pages

vimleshtna:


thanhz (approved):
Looks good, thanks for fixing

PR #632, opened 2023-06-16
TDR-3244 update series consignment status

thanhz:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -117,15 +117,10 @@ class ConsignmentService(
   }
 
   def updateSeriesIdOfConsignment(updateConsignmentSeriesIdInput: UpdateConsignmentSeriesIdInput): Future[Int] = {
-    addSeriesStatus(updateConsignmentSeriesIdInput.consignmentId)
+    consignmentStatusRepository.updateConsignmentStatus(updateConsignmentSeriesIdInput.consignmentId, "Series", Completed, Timestamp.from(timeSource.now))
```
TomJKing, line 120: Should probably not be updating the status until the series id has been added to the consignment. Would change this method to something like this:
  ```
def updateSeriesIdOfConsignment(updateConsignmentSeriesIdInput: UpdateConsignmentSeriesIdInput): Future[Int] = {
    for {
      result <- consignmentRepository.updateSeriesIdOfConsignment(updateConsignmentSeriesIdInput)
      seriesStatus = if (result == 1) { Completed } else { Failed }
      _ <- consignmentStatusRepository.updateConsignmentStatus(updateConsignmentSeriesIdInput.consignmentId, "Series", seriesStatus, Timestamp.from(timeSource.now))

    } yield result
  }
```

File: src/test/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentServiceSpec.scala [hunk 2]
```diff
@@ -586,36 +586,6 @@ class ConsignmentServiceSpec extends AnyFlatSpec with MockitoSugar with ResetMoc
     exception.getMessage should equal("Existing consignment upload status is 'InProgress', so cannot start new upload")
   }
 
-  "addSeriesStatus" should "update the consignment status for a series" in {
```
TomJKing, line 589: The `"updateSeriesIdOfConsignment" should "update the seriesId for a given consignment"` should be updated to test whether the status gets updated in the same call.

TomJKing (commented):
Looks good. Couple of minor comments. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
thanhz: This makes sense. I've made the changes

TomJKing (approved):
Looks good. Thanks

PR #637, opened 2023-06-19
Use consignment id arg instead of value from file

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-transfer-frontend/pull/3094

To remove the need to make an extra call to the DB use the consignment id arg passed into the mutation

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
@@ -72,9 +72,9 @@ class FileMetadataService(
 
   def deleteFileMetadata(input: DeleteFileMetadataInput, userId: UUID): Future[DeleteFileMetadata] = {
     val propertiesToDelete = descriptionDeletionHandler(input.propertyNames)
+    val consignmentId: UUID = input.consignmentId.getOrElse(throw InputDataException("No consignment id"))
+    val fileIds: Set[UUID] = input.fileIds.toSet
     for {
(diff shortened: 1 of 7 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -72,9 +72,9 @@ class FileMetadataService(
 
   def deleteFileMetadata(input: DeleteFileMetadataInput, userId: UUID): Future[DeleteFileMetadata] = {
     val propertiesToDelete = descriptionDeletionHandler(input.propertyNames)
+    val consignmentId: UUID = input.consignmentId.getOrElse(throw InputDataException("No consignment id"))
```
vimleshtna, line 75: Can we make `consignmentId` mandatory instead of checking here if it is `None` and then throwing an exception?

https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/graphql/fields/FileMetadataFields.scala#L40

vimleshtna (commented):
Looks good, just asked a question.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
TomJKing: Not without causing a breaking change, which we need to avoid.

vimleshtna: OK

vimleshtna (approved):

PR #664, opened 2023-08-01
TDR-2628 Feature block http4s

thanhz:
This pr will enable us to run the consignment application using Http4s or Akka http. This is so that we can test http4s. If all goes well. We will decommission Akka.


src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServerHttp4s.scala
```diff
@@ -0,0 +1,156 @@
...
+import uk.gov.nationalarchives.tdr.api.consignmentstatevalidation.{ConsignmentStateException, ConsignmentStateValidator}
+import uk.gov.nationalarchives.tdr.api.db.repository._
+import uk.gov.nationalarchives.tdr.api.graphql.DataExceptions.InputDataException
+import uk.gov.nationalarchives.tdr.api.graphql.{ConsignmentApiContext, DeferredResolver, ErrorCodes, GraphQlTypes}
+import uk.gov.nationalarchives.tdr.api.service._
+import uk.gov.nationalarchives.tdr.keycloak.Token
+
+import scala.concurrent.ExecutionContext
+
+class GraphQLServerHttp4s() {
```
TomJKing, line 23: There is a lot of common code between this new class and the existing `GraphQlServer.class`.

To avoid this duplication think its worth defining a new trait that contains the common code, and have the two classes inherit from the trait. 

Something like this:

```
trait Server { 
  def handleException(...): HandledException  = { ... }
  ... etc ...
}
```

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServerHttp4s.scala
```diff
@@ -0,0 +1,156 @@
...
+            deferredResolver = new DeferredResolver,
+            middleware = new ValidationAuthoriser :: new ConsignmentStateValidator :: Nil,
+            exceptionHandler = exceptionHandler
+          )
+        )
+      ).recover { case error: QueryAnalysisError =>
+        error.resolveError
+      }
+
+      //        .map(OK -> _).recover {
```
TomJKing, line 145: Is this meant to be here?

TomJKing (commented):
Looks good. Just some comments about reducing the amount of duplicate code. This is a temporary change, as we will decommission the Akka stuff, but I think its worth reducing the duplicate code, as these changes might be in for a while, whilst testing is done.

We should probably ask Sam to take a look as well.

src/main/scala/uk/gov/nationalarchives/tdr/api/http/GraphQLServerHttp4s.scala
thanhz: Nope I've removed it

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Http4sServer.scala
```diff
@@ -0,0 +1,86 @@
...
+          `Content-Type`(MediaType.application.json)
+        )
+    case _ @GET -> Root / "healthcheck" => Ok()
+    case _ @GET -> Root / "healthcheck-full" =>
+      IO.fromFuture(IO(fullHealthCheck.checkDbIsUpAndRunning(DbConnectionHttp4s().db)))
+        .flatMap(_ => Ok())
+  }
+
+  def corsWrapper: Http[IO, IO] = CORS.policy
+    .withAllowOriginHost(Set(Origin.Host(Uri.Scheme.http, Uri.RegName("localhost"), Option(9001))))
```
TomJKing, line 56: Should the `localhost` be hardcoded here? Would have thought for TDR environments it would be different, so maybe this should be in the application conf?

TomJKing (commented):
Looks good. Just one thing I've noticed. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/http/Http4sServer.scala
thanhz: hmm good spot.. I'm not actually sure what the value is supposed to be in here @MancunianSam any suggestions?

TomJKing: @thanhz I think it needs to be the domain?

MancunianSam: I think it does but I'm just trying to test it and it doesn't matter what you put in there, it lets the request through and you don't get an access control header. I think I'm missing something.

MancunianSam:
The other thing I can think of is that it's not logging the requests to the routes. You can add some basic logging of the routes by doing something like:
`val loggingWrapper: HttpApp[IO] = httpApp[IO](true, false)(corsWrapper)`

Then in the server, call that instead of corsWrapper
```
    .withPort(port"8080")
    .withHttpApp(loggingWrapper)
    .build
```
It might help a little bit and it might not be good enough when you're using this exclusively but it might be worth it?

MancunianSam (approved):
Looks good, cheers.

PR #668, opened 2023-08-09
Setting Akka version to last open source version

TomJKing:


thanhz (commented):
Do we not need to downgrade akkaHttpVersion down to 10.1.2 or earlier?

TomJKing:
> Do we not need to downgrade akkaHttpVersion down to 10.1.2 or earlier?

@thanhz No idea. Not sure this is covered by the licence change.

thanhz (approved):

PR #706, opened 2023-10-11
Add seriesName too if user passed a seriesId while calling addConsignment

vimleshtna:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -172,6 +173,14 @@ class ConsignmentService(
       .map(cr => convertRowToConsignment(cr))
       .map(c => ConsignmentEdge(c, c.consignmentReference))
   }
+
+  private def getSeriesName(seriesId: Option[UUID]): Future[Option[String]] = {
```
TomJKing, line 177: Should use this method in the `updateSeriesOfConsignment` method to reduce duplicate code.

TomJKing (commented):
Looks good, thanks. Just one minor comment.

TomJKing (approved):
Looks good. Thanks

PR #711, opened 2023-10-13
TDR-3430 - Remove series field from getConsignment query

vimleshtna:
Remove `series` field from the `getConsignment` query
We can't remove `transferringBody` as we are still using `transferringBody.code` in `tdr-reporting`


TomJKing (approved):
Looks good. Thanks

PR #716, opened 2023-10-17
TDR-3499 Populate file table with parent reference

thanhz:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -54,15 +53,9 @@ class FileService(
...
-      val allNodesWithReference = if (referenceGeneratorFeatureBlock) {
-        allNodes.keys.zip(allNodes.values.zip(List.fill(allNodes.size)(None))).toMap
-      } else {
-        val generatedReferences = referenceGeneratorService.getReferences(allNodes.size)
-        allNodes.keys.zip(allNodes.values.zip(generatedReferences.map(Some(_)))).toMap
-      }
-      (allNodesWithReference map { case (path, (treeNode, reference)) =>
+      ((allEmptyDirectoryNodes ++ allFileNodes) map { case (path, treeNode) =>
         val parentId = treeNode.parentPath.map(path => allFileNodes.getOrElse(path, allEmptyDirectoryNodes(path)).id)
+        val parentFileReference = treeNode.parentPath.flatMap(path => allFileNodes.getOrElse(path, allEmptyDirectoryNodes(path)).reference)
```
TomJKing, line 58: I would get the whole parent node first, then retrieve its id and reference separately. Something like this:

`val parentNode: Option[TreeNode] = treeNode.parentPath.map(path => allFileNodes.getOrElse(path, allEmptyDirectoryNodes(path)))`

Saves searching all the nodes twice to get the parent node.

TomJKing (commented):
Looks good. Just one comment.

Additionally with this PR the "parent reference" is not being returned in queries that return file metadata. This case class will need to be updated: https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala#L184 with the relevant graphql queries being updated to return the parent reference.

This would be the graphql query that would need to include the parent reference to it gets into the Bagit package: https://github.com/nationalarchives/tdr-generated-graphql/blob/master/src/main/graphql/GetConsignmentExport.graphql

Not sure if you were intending to do this change as part of a separate PR?

src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
thanhz: Good idea, I've made the changes.

thanhz:
As suggested will make a separate pr for returning of parentReference for file metadata queries

TomJKing (approved):
Looks good. Thanks.

PR #718, opened 2023-10-24
Add fileReference and parentReference to fileMetadata

thanhz:
FileReference and ParentReference needs to be inserted into the FileMetadata table in order to included in the export bagit.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
@@ -138,6 +138,8 @@ object FileMetadataService {
   val ClosureStartDate = "ClosureStartDate"
   val Filename = "Filename"
   val FileType = "FileType"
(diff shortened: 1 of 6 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileMetadataService.scala
```diff
@@ -138,6 +138,8 @@ object FileMetadataService {
   val ClosureStartDate = "ClosureStartDate"
   val Filename = "Filename"
   val FileType = "FileType"
+  val FileReference = "FileReference"
```
TomJKing, line 141: Realise this is nothing to do with your PR. But could you add a unit test to check all these static values in `FileMetadataService` object? Good practice in case someone inadvertently changes them, at least they'll be a unit test warning.

TomJKing (commented):
Looks good. Just one minor comment. Thanks

TomJKing (approved):
Looks good. Thanks

PR #719, opened 2023-10-27
TDR-3504 update rds certificate

thanhz:
Download rds certificate bundle from here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesAllRegions


TomJKing (approved):
Looks good. Thanks

Suggest running the same checks as with the tdr-statuses update to make sure this still works with the DB using the old certificate.

PR #722, opened 2023-11-10
Tdr 3566 accommodate large consignment total file size

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/FileMetadataRepository.scala
```diff
@@ -20,15 +20,15 @@ class FileMetadataRepository(db: Database)(implicit val executionContext: Execut
   private val insertFileStatusQuery =
```
thanhz, line 20: `InsertFileStatusQuery` doesn't seem to be used. The class also has a few unused imports that can be removed.

thanhz (commented):
Looks good, just some minor points

thanhz (approved):

PR #723, opened 2023-11-13
TDR-3414 - Use metadata validation library to validate metadata

vimleshtna:
Use metadata validation library to validate while updating/deleting additional metadata
Add a feature access block for the library

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
@@ -32,7 +32,7 @@ class DisplayPropertiesService(displayPropertiesRepository: DisplayPropertiesRep
 
   def getDisplayProperties: Future[Seq[DisplayPropertyField]] = {
     for {
-      displayProperties <- displayPropertiesRepository.getDisplayProperties
+      displayProperties <- displayPropertiesRepository.getDisplayProperties()
     } yield {
       toDisplayProperty(displayProperties)
     }
@@ -40,7 +40,7 @@ class DisplayPropertiesService(displayPropertiesRepository: DisplayPropertiesRep
 
   def getActiveDisplayPropertyNames: Future[Seq[String]] = {
     for {
-      displayProperties <- displayPropertiesRepository.getDisplayProperties("Active", "true")
+      displayProperties <- displayPropertiesRepository.getDisplayProperties(Some("Active"), Some("true"))
     } yield {
       displayProperties.flatMap(_.propertyname)
     }
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
@@ -37,4 +37,12 @@ class DisplayPropertiesService(displayPropertiesRepository: DisplayPropertiesRep
       toDisplayProperty(displayProperties)
     }
   }
+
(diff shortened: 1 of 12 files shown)


thanhz (commented):
Looks good, just a few comments

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ValidateFileMetadataService.scala
```diff
@@ -123,3 +198,37 @@ class ValidateFileMetadataService(
...
+
+    val closureFields = fields.filter(f => f.name == ClosureType)
+    val descriptiveFields = fields.filter(f => f.propertyGroup.contains("OptionalMetadata"))
+
+    val closureMetadataCriteria = createCriteria(closureFields, fields).head
+    val descriptiveMetadataCriteria = createCriteria(descriptiveFields, fields)
+    new MetadataValidation(closureMetadataCriteria, descriptiveMetadataCriteria)
+  }
+
+  private def createCriteria(customMetadata: List[CustomMetadataField], allCustomMetadata: List[CustomMetadataField]): List[MetadataCriteria] = {
```
TomJKing, line 214: This function is not optimised for tail recursion. Should probably look into making it tail recursive as potentially could get stack overflow issues.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/DisplayPropertiesService.scala
```diff
@@ -37,4 +37,12 @@ class DisplayPropertiesService(displayPropertiesRepository: DisplayPropertiesRep
       toDisplayProperty(displayProperties)
     }
   }
+
+  def getActiveDisplayPropertyNames: Future[Seq[String]] = {
```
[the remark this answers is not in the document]
vimleshtna: Initially, I thought the same but then I created an overloaded method. Anyway, it doesn't really make any difference so I changed the existing method as you said, thanks.

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ValidateFileMetadataService.scala
```diff
@@ -123,3 +198,37 @@ class ValidateFileMetadataService(
...
+
+object MetadataValidationUtils {
+
+  def createMetadataValidation(fields: List[CustomMetadataField]): MetadataValidation = {
+
+    val closureFields = fields.filter(f => f.name == ClosureType)
+    val descriptiveFields = fields.filter(f => f.propertyGroup.contains("OptionalMetadata"))
+
+    val closureMetadataCriteria = createCriteria(closureFields, fields).head
+    val descriptiveMetadataCriteria = createCriteria(descriptiveFields, fields)
```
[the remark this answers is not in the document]
vimleshtna: Actually, ClosureType is the only metadata at the top level rest of them are dependencies. So, we are returning only head

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ValidateFileMetadataService.scala
vimleshtna: I don't think it is possible with the metadata criteria because we are recalling the same function over values and settings as dependencies. Also, the custom metadata is coming from the DB which is properly structured. So, I don't think there will be any stack overflow issues.

TomJKing: Fair point.

thanhz (approved):
Looks good. Thanks

PR #738, opened 2024-02-23
Provide a dummy Akka token value as a default

TomJKing:
Scala Steward is failing because an exception was thrown if no akka token value

There is no way to pass the akka token into Scala Steward so providing a dummy value for the token instead of throwing the exception


thanhz (approved):
Looks good, thanks for doing this.

PR #799, opened 2024-04-05
Allow backend check client to access any consignment and files

vimleshtna:
Draft metadata validator doesn't know the user id and due to this, the `addOrUpdateBulkFileMetadata` mutation throws an error that `User 'id' does not own the files they are trying to access`. So, for now, allow backend check client to access any consignment and files.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/metadatainputvalidation/MetadataInputTags.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/metadatainputvalidation/MetadataInputTags.scala
@@ -6,6 +6,7 @@ import uk.gov.nationalarchives.tdr.api.graphql.fields.FileMetadataFields.{AddOrU
 import uk.gov.nationalarchives.tdr.api.graphql.{ConsignmentApiContext, ValidationTag}
 import uk.gov.nationalarchives.tdr.api.model.file.NodeType
 import uk.gov.nationalarchives.tdr.api.auth.AuthorisationException
+import uk.gov.nationalarchives.tdr.api.auth.ValidateBody.exportRole
 import uk.gov.nationalarchives.tdr.api.graphql.DataExceptions.InputDataException
 import uk.gov.nationalarchives.tdr.api.service.FileService.FileDetails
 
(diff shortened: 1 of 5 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/metadatainputvalidation/MetadataInputTags.scala
```diff
@@ -25,19 +26,21 @@ case class ValidateMetadataInput[T](argument: Argument[T]) extends MetadataInput
...
+    // TODO: Should use a new client for the draft metadata upload https://national-archives.atlassian.net/browse/TDRD-181
+    val draftMetadataValidatorAccess = token.backendChecksRoles.contains(exportRole)
 
     for {
       fileFields <- ctx.ctx.fileService.getFileDetails(inputFileIds)
       nonOwnership = fileFields.exists(_.userId != userId)
     } yield {
       nonOwnership match {
-        case true => throw AuthorisationException(s"User '$userId' does not own the files they are trying to access")
+        case true if !draftMetadataValidatorAccess => throw AuthorisationException(s"User '$userId' does not own the files they are trying to access")
```
TomJKing, line 43: Maybe refactor this a little to simplify. Something like this:
```
...
for {
      fileFields <- ctx.ctx.fileService.getFileDetails(inputFileIds)
      noAccess = fileFields.exists(_.userId != userId) || !draftMetadataValidatorAccess
    } yield {
      noAccess match {
        case true => throw AuthorisationException(s"Access denied to file metadata")
        case _ if inputErrors(fileFields, inputFileIds, inputConsignmentId) =>
          throw InputDataException("Input contains directory id or contains non-existing file id or consignment id is incorrect")
        case _ => continue
      }
    }
...
```

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileMetadataRouteSpec.scala
```diff
@@ -466,6 +466,38 @@ class FileMetadataRouteSpec extends TestContainerUtils with Matchers with TestRe
     checkNoFileMetadataAdded(utils, "property2")
   }
 
+  "addOrUpdateBulkFileMetadata" should "add or update all file metadata when consignment belongs to another user but it got the backend check token" in withContainers {
```
TomJKing, line 469: Maybe reword test to: `... when called by backend checks client" ...`

src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileMetadataRouteSpec.scala
```diff
@@ -466,6 +466,38 @@ class FileMetadataRouteSpec extends TestContainerUtils with Matchers with TestRe
     checkNoFileMetadataAdded(utils, "property2")
```
TomJKing, line 466: The other tests will need to be re-worded to make it explicit that the "Not Authorised" exception is thrown when the user doesn't own the file and the backend checks client is not used.

TomJKing (commented):
Looks good. Thanks. Couple of minor comments.

TomJKing (approved):
Looks good. Thanks

PR #801, opened 2024-04-09
Adding run-name to deploy script

Tom-Hallett:
Minor change to output more meaningful name to the workflow pages


TomJKing (approved):
Looks good. Thanks for putting this in.

PR #849, opened 2024-06-12
TDRD-217 Overwrite all values

thanhz:
The draft-metadata lambda can pass in properties with empty values in this case we want to delete empty values from the `fileMetadata` table but ensure we don't add empty values to the db.

Added new `propertyProtected` to `CustomMetadata` to allow us to filter on properties that shouldn't be editable i.e. UUID. Will thrown an exception if trying to add metadata that are protected.


Tom-Hallett (approved):
Looks good, thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/service/CustomMetadataPropertiesService.scala
```diff
@@ -88,7 +89,8 @@ class CustomMetadataPropertiesService(customMetadataPropertiesRepository: Custom
       metadataValues.toList,
       fp.uiordinal.getOrElse(Int.MaxValue),
       fp.allowexport,
-      fp.exportordinal.map(_.toInt)
+      fp.exportordinal.map(_.toInt),
+      Set(MandatoryMetadata, SystemProperty).contains(fp.propertygroup.get)
```
vimleshtna, line 93: Can you create a variable for the protected field in class and then use it here?

`val protectedPropertyGroups = Set(MandatoryMetadata, SystemProperty)`

vimleshtna (commented):
Looks good. Just a minor comment. Thanks

vimleshtna (approved):
Looks good. Thanks

PR #852, opened 2024-06-13
TDRD-148: increase server timeouts on staging to confirm issue

MarkKingsbury:
large update (10,000 files) fails with ` Request timeout encountered for request [POST /graphql` in consignment-api so increasing to confirm that found the bottleneck / reveal next errors behind this..


src/main/resources/application.staging.conf
```diff
@@ -1 +1,8 @@
 include "application.env-base"
+
+akka.http {
+  server {
+    idle-timeout = 900 s
+    request-timeout = 720 s
+  }
+}
```
vimleshtna, line 8: Missing a new line in the end of the file

vimleshtna (commented):
The `request-timeout` and `idle-timeout` are very high. I think let's try it first and if it works then we can try doubling the existing timeouts:

```
akka.http {
  server {
    idle-timeout = 360 s
    request-timeout = 240 s
  }
}

MarkKingsbury:
thanks @vimleshtna  - added that new line & agree we'll prove here that got to the bottle neck - or see more about what is behind it... and I'll also be able to get some estimates of how long it's taking so we can decide if we want to bump to allow that or look at refactor to improve speed.

PR #853, opened 2024-06-14
TDRD-148: bump future time-out in validate

MarkKingsbury:
As a follow-up to yesterday increase on `staging` of request-timeout  https://github.com/nationalarchives/tdr-consignment-api/pull/852 to understand where bottle neck is:
- that has stopped `request-timeout` error previously seen but we encounter a (new) 504 behind it. 
- think this `await result` will be source of that... mainly based on https://github.com/nationalarchives/tdr-consignment-api/commit/cd3aec131042fb145f3aadaa627000f5b8d66b15 where it was updated alongside the request-timeout in 2022 (and noted at that time that going to 2 mins worked with 10,000 files so testing to see if a little more time is all that is required).


vimleshtna (approved):
Looks good. Thanks

PR #874, opened 2024-07-02
TDRD-227 - add MetadataReview status type

vimleshtna:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
```diff
@@ -115,8 +115,19 @@ class ConsignmentStatusService(
...
+      "Series",
+      "TransferAgreement",
+      "Upload",
+      "ClientChecks",
+      "DraftMetadata",
+      "ClosureMetadata",
+      "DescriptiveMetadata",
+      "ConfirmTransfer",
+      "Export",
+      "MetadataReview"
```
thanhz, line 129: I believe the status is supposed to be `DTAReview`

thanhz (commented):
Looks good

src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentStatusService.scala
vimleshtna: Actually, we call it now "Metadata Review" so I used it as `MetadataReview` type and it makes more sense to have as `MetadataReview` type because all the pages are called Metadata Review.

thanhz (approved):

PR #875, opened 2024-07-02
Add new `getConsignmentsForReview` query

thanhz:
A new query for fetching all consignments that have `DtaReview` status `InProgress`. Query can only be called by `TNAUser` type.

### DIFF
--- a/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
+++ b/src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
@@ -59,6 +59,16 @@ class ConsignmentRepository(db: Database, timeSource: TimeSource) {
     db.run(query.result)
   }
 
+  def getConsignmentsForReview: Future[Seq[ConsignmentRow]] = {
+    val query = Consignment
+      .join(Consignmentstatus)
+      .on(_.consignmentid === _.consignmentid)
+      .filter(_._2.statustype === "DTAReview")
+      .filter(_._2.value === "InProgress")
+      .map(_._1)
+    db.run(query.result)
+  }
+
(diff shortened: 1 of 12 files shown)


src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
```diff
@@ -6,6 +6,7 @@ import slick.jdbc.PostgresProfile.api._
 import uk.gov.nationalarchives.Tables.{Body, BodyRow, Consignment, ConsignmentRow, Consignmentstatus, ConsignmentstatusRow, File, Series, SeriesRow}
 import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields
 import uk.gov.nationalarchives.tdr.api.graphql.fields.ConsignmentFields.{ConsignmentFilters, StartUploadInput}
+import uk.gov.nationalarchives.tdr.api.service.FileStatusService.InProgress
```
TomJKing, line 9: Should really move the Status values and type to a separate trait / class instead of using values from the `FileStatusService`.

TomJKing (commented):
Looks good. Need to add unit tests for the new endpoint.

vimleshtna (commented):
Looks good. Just a few minor comments. Thanks

src/main/scala/uk/gov/nationalarchives/tdr/api/db/repository/ConsignmentRepository.scala
thanhz: I've created a new `Statuses` object in the util directory

TomJKing (approved):
Looks good. Thanks.

Should refactor the code now have the statuses separated out and remove the usage of the `FileStatusService` statuses.

Looking at the code quickly they're used in:

-  `ConsignmentService`
-  `ConsignmentStatusService`
-  `FileMetadataService`
-  `FileService`
-  `FileStatusService`
-  `ValidateFileMetadataService`

This will be quite a lot of changes so it should be split out into a new branch/PR.

PR #918, opened 2024-09-02
TDRD-425 - Allow transfer adviser user only to update the consignment status

vimleshtna:


TomJKing:
@MarkKingsbury Would you mind taking a look at this for Vimlesh? Thanks

annielh (commented):
I think because the `ValidateUserHasAccessToConsignment` tag controls all access to a consignment, read and write, this would stop a metadata viewer user being able to fetch any of the information required to render the pages? I think a little more refactoring and/or duplication will be needed to get the granularity we want. Let me know if you want to chat through.

We also want the update and view permissions to extend to DTA Reviewers and Metadata Viewers only for consignments that have a `MetadataReviewStatus` of `InProgress` (and come to think of it, perhaps also `CompletedWithIssues`)

PR #1021, opened 2025-01-27
TDRD-181 Use new draftMetadata client

thanhz:


vimleshtna (approved):
Looks good. Thanks.

Can you update the draft-metadata-validator as well to use the draft metadata instead of backend checks client secret

PR #1027, opened 2025-02-04
update akkaHttpVersion as this is compatible with akkaVersion 2.10.0

ian-hoyle:
hopefully this will fix the slick upgrade. Quite a few changes including in pre-release
chore: drop usage of akka.dispatch.ExecutionContexts in https://github.com/akka/akka-http/pull/4422


vimleshtna (approved):
Looks good. Thanks

PR #1106, opened 2025-04-14
Tdrd 718 consignment metadata service

TomJKing:


thanhz (approved):
Looks good to me :+1:  Will be interesting to see what happens if you try to save a file with unusual characters.

PR #1150, opened 2025-05-23
Sptdr service transfer client user

TomJKing:
To support SharePoint and other third party service transfers the adding of File entries to the Database will be done via a service client user

The user id in the bearer access token will be the service client user and not the user that owns the transfer this will cause a discrepency in the database when creating initial transfer file entries

The change will allow the user id of the user who owns the transfer to be set in the database entries along with checking that this user id matches the user id for the consignment


thanhz (approved):
Looks good to me :+1:

PR #1173, opened 2025-06-04
Update spec for new mockito

ian-hoyle:


build.sbt
```diff
@@ -63,35 +63,35 @@ libraryDependencies ++= Seq(
...
   "io.github.hakky54" % "logcaptor" % "2.11.0" % Test,
   "com.dimafeng" %% "testcontainers-scala-scalatest" % testContainersVersion % Test,
   "com.dimafeng" %% "testcontainers-scala-postgresql" % testContainersVersion % Test,
   "com.github.tomakehurst" % "wiremock-standalone" % "3.0.1" % Test
 )
 
 //Akka Alpakka requires slick v3.5.2
-dependencyOverrides += "com.typesafe.slick" %% "slick" % "3.5.2"
-dependencyOverrides += "com.typesafe.slick" %% "slick-hikaricp" % "3.5.2"
+dependencyOverrides += "com.typesafe.slick" %% "slick" % "3.6.1"
```
TomJKing, line 93: Is it safe to upgrade the Slick versions now? If so these dependency overrides should be removed now as they're using the latest version of Slick anyway.

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
```diff
@@ -5,7 +5,8 @@ import com.typesafe.config.ConfigFactory
 import org.mockito.ArgumentMatchers._
 import org.mockito.stubbing.ScalaOngoingStubbing
 import org.mockito.{ArgumentCaptor, ArgumentMatchers, MockitoSugar}
-import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.concurrent.{Eventually, ScalaFutures}
```
TomJKing, line 8: Unused imports:

- `import org.scalatest.concurrent.{Eventually, ...`
- `import org.scalatest.events.Event`

TomJKing (commented):
Looks good. Couple of minor comments. Thanks

build.sbt
ian-hoyle: binary incompatibles are resolved by using 3.6.1 but there is no reference to Akka support above 5.2 in latest release so reverted back to 3.5.2

src/test/scala/uk/gov/nationalarchives/tdr/api/service/FileServiceSpec.scala
ian-hoyle: removed

TomJKing (dismissed):
Looks good. Thanks.

PR #1174, opened 2025-06-05
TDRD-795 Remove additional metadata related code

thanhz:
- Remove statuses 'DescriptiveMetadata', ClosureMetadata, `NotEntered` and `Incomplete`
- Remove old validation code
- Remove DisplayProperties code
- Remove getAllDescendant
- Remove updatebulkmetadata & deleteFileMetadata


TomJKing (commented):
Hi,

I've added some initial comments, but have not fully reviewed.

- the fields that are being deleted are all still referenced in the generated graphql, so deleting them will be a breaking change.
- the other repos affected should be cleaned up first before under taking the changes to the API.
- also the PR should be split into smaller ones, as its too big to review in a single go

Given the above I would suggest dealing with each field to be deleted as a separate task:
1. clean up the repos which use the field, ensuring it will not be a breaking change
2. remove the query/mutation from the generated graphql
3. delete the fields and attendant services etc from the consignment api

PR #1184, opened 2025-06-10
TDRD-795-remove-getallDescendents

thanhz:
This was used to display the tree view of files in the UI which has been removed from the frontend.


src/test/scala/uk/gov/nationalarchives/tdr/api/routes/FileRouteSpec.scala
```diff
@@ -23,7 +23,6 @@ class FileRouteSpec extends TestContainerUtils with Matchers with TestRequest {
   case class FileNameAndPath(fileName: String, path: String)
```
TomJKing, line 23: There are a couple of case classes that refer to "all descendants" that should be removed:
- line 30: `case class GraphqlQueryDataAllDescendants`;
- line 37: `case class AllDescendants`

PR #1198, opened 2025-06-16
TDRD-795 Remove skipValidation argument from AddOrUpdateFileMetadata

thanhz:
Will deploy new version of graphql dependency once this is merged


TomJKing (commented):
Looks good. Thanks

Did you test locally whether the generated graphql created with this changes will break any client's using the removed `skipValidation` parameter. Not sure if it would be a breaking change?

thanhz:
> Looks good. Thanks
> 
> Did you test locally whether the generated graphql created with this changes will break any client's using the removed `skipValidation` parameter. Not sure if it would be a breaking change?

I've prepared a couple prs to remove the `skipValidation` argument from clients that use it. I think the changes might just have to go in quickly to prevent downtime.

- https://github.com/nationalarchives/tdr-e2e-tests/pull/1056
- https://github.com/nationalarchives/tdr-draft-metadata-validator/pull/561

TomJKing (approved):
Looks good. Thanks. 

Make sure you inform the team this is a breaking change and the env will be potentially broken until the calling clients are updated.

PR #1382, opened 2026-04-07
[TDRD 1216] Give client with notify export details role permission to access consignments

annielh:


vimleshtna (approved):
Looks good. Thanks

PR #1388, opened 2026-04-14
Tdrd 1330 metadata review

ian-hoyle:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala
```diff
@@ -25,13 +27,22 @@ class ConsignmentService(
...
     transferringBodyService: TransferringBodyService,
+    metadataReviewLogRepository: MetadataReviewLogRepository,
     timeSource: TimeSource,
     uuidSource: UUIDSource,
     config: Config
 )(implicit val executionContext: ExecutionContext) {
 
   val maxLimit: Int = config.getInt("pagination.consignmentsMaxLimit")
 
+  // Review status sort ordering: Requested > Rejected > Approved > Completed
```
annielh, line 38: Should this live here or in where the enum is defined in the commons? Would it ever be used elsewhere, ie the frontend client?

annielh (approved):

PR #1400, opened 2026-04-28
Replace local Statuses enums with common-utils imports

annielh:
## Summary

Migrates status type and value references from the local `Statuses.scala` to `tdr-common-utils` enums.

## Changes

- **10 test spec files**: replaced string literals (`"Upload"`, `"InProgress"`, etc.) with common-utils enum references (`UploadType.id`, `InProgressValue.value`, etc.)
- **2 production files**: migrated `ConsignmentRepository` and `ConsignmentStatusService` imports from local to common-utils
- **Slimmed `Statuses.scala`**: reduced to only `ClosureMetadataType` and `DescriptiveMetadataType`, which are not yet in common-utils

## Note

`ClosureMetadataType` and `DescriptiveMetadataType` may no longer be needed following the transition to the CSV metadata upload route. A TODO has been added to investigate and either remove them or add them to common-utils.


src/main/scala/uk/gov/nationalarchives/tdr/api/utils/Statuses.scala
```diff
@@ -1,94 +1,10 @@
 package uk.gov.nationalarchives.tdr.api.utils
 
+// Legacy status types not yet in tdr-common-utils.
```
TomJKing: Think its safe to delete these. Checked the prod database and they're not used anywhere that I can see.

TomJKing (commented):
Looks good. Thanks.

There is a ticket in the backlog for the this move: https://national-archives.atlassian.net/browse/TDRD-1338

Have all the statuses been migrated as part of this PR, if so that ticket can be marked as done.

TomJKing (approved):
Looks good. Thanks

PR #1429, opened 2026-07-07
Generate and persist asset id in database

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-consignment-export/pull/1963

DR2 has concept of an asset to support where there are multiple digital objects associated with a single record.

This is modelled using an asset id.

To ensure idempotency TDR should generate and add the asset id when creating entries for uploaded digital objects

Current process on export is to use the TDR 'file id' as the asset id and generate an ephemeral id for the digital object id (file id)

On a re-export this causes DR2 errors because the ephemeral id is re-generated and not persisted


vimleshtna (approved):
Looks good. Thanks

PR #1434, opened 2026-07-15
Assign asset id feature access block

TomJKing:


src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileService.scala
```diff
@@ -70,7 +71,7 @@ class FileService(
         val parentId = parentNode.map(_.id)
         val parentFileReference = parentNode.flatMap(_.reference)
         val fileId = treeNode.id
-        val assetId = uuidSource.uuid
+        val assetId = if (blockAssignAssetId) None else Some(uuidSource.uuid)
```
vimleshtna, line 74: Can you add a test for if `blockAssignAssetId` is `true` then the value should be `None` for `assetId`?

vimleshtna (commented):
Looks good. Just a minor comment. Thanks

vimleshtna (approved):
Looks good. Thanks

